PR #42, opened 2023-06-16
Separate computation of recall [skip ci]

lums658:
This PR moves the computation of recall to three overloaded functions in `stats.h`.  They all compute the recall value between a given groundtruth and the computed top k nearest neighbors.
```
// Compute recall using `tiledb::Context` and `groundtruth_uri` to get recall
template <class Topk>
double compute_recall(const tiledb::Context ctx, const std::string& groundtruth_uri, const Topk& top_k);

// Compute recall using `groundtruth_uri` with default `tiledb::Context` to get recall
template <class Topk>
double compute_recall(const std::string& groundtruth_uri, const Topk& top_k);

// Compute recall with an already opened ground truth `tdbMatrix`
template <class Matrix, class Topk>
double compute_recall(const Matrix& groundtruth, const Topk& top_k);


lums658:
Implemented `count_intersections()` and use that for computing recall.

PR #52, opened 2023-06-19
Log stats on array openings and query submissions.

teo-tsirpanis:
[SC-30470](https://app.shortcut.com/tiledb-inc/story/30470/add-stats-logging-for-tiledb-operations-in-tdbmatrix-tdbcolmajormatrix)

### DIFF
--- a/apis/python/src/tiledb/vector_search/module.cc
+++ b/apis/python/src/tiledb/vector_search/module.cc
@@ -14,6 +14,8 @@ using Ctx = tiledb::Context;
 bool global_debug = true;
 double global_time_of_interest;
 
+bool enable_stats = false;
+
 PYBIND11_MAKE_OPAQUE(std::vector<uint32_t>);
 PYBIND11_MAKE_OPAQUE(std::vector<uint64_t>);
 
@@ -248,6 +250,16 @@ PYBIND11_MODULE(_tiledbvspy, m) {
          const ColMajorMatrix<int32_t>& ground_truth) -> bool {
         return validate_top_k(top_k, ground_truth);
       });
+
+  m.def("enable_stats", []() {
+    enable_stats = true;
+    tiledb::Stats::enable();
+  });
+
+  m.def("disable_stats", []() {
+    enable_stats = false;
+    tiledb::Stats::disable();
+  });
 
   declare_kmeans_query<uint8_t>(m, "u8");
   declare_kmeans_query<float>(m, "f32");
--- a/apis/python/src/tiledb/vector_search/module.cc
+++ b/apis/python/src/tiledb/vector_search/module.cc
@@ -15,7 +15,7 @@ bool global_debug = true;
 double global_time_of_interest;
 
 bool enable_stats = false;
-FILE* stats_file = stdout;
+std::vector<json> core_stats;
 
 PYBIND11_MAKE_OPAQUE(std::vector<uint32_t>);
 PYBIND11_MAKE_OPAQUE(std::vector<uint64_t>);
@@ -252,14 +252,22 @@ PYBIND11_MODULE(_tiledbvspy, m) {
         return validate_top_k(top_k, ground_truth);
       });
 
-  m.def("enable_stats", []() {
+  m.def("stats_enable", []() {
     enable_stats = true;
     tiledb::Stats::enable();
   });
 
-  m.def("disable_stats", []() {
+  m.def("stats_disable", []() {
     enable_stats = false;
     tiledb::Stats::disable();
+  });
+
+  m.def("stats_reset", []() {
+    core_stats.clear();
+  });
+
+  m.def("stats_dump", []() {
+    return json{core_stats}.dump();
   });
 
   declare_kmeans_query<uint8_t>(m, "u8");
--- a/apis/python/src/tiledb/vector_search/module.cc
+++ b/apis/python/src/tiledb/vector_search/module.cc
@@ -15,6 +15,7 @@ bool global_debug = true;
 double global_time_of_interest;
 
 bool enable_stats = false;
+FILE* stats_file = stdout;
 
 PYBIND11_MAKE_OPAQUE(std::vector<uint32_t>);
 PYBIND11_MAKE_OPAQUE(std::vector<uint64_t>);
--- a/src/include/stats.h
+++ b/src/include/stats.h
@@ -56,19 +56,21 @@ using json = nlohmann::json;
 // Make stats support opt-in to avoid requiring to define an enable_stats variable on all projects.
 #ifdef TILEDBVS_ENABLE_STATS
 extern bool enable_stats;
-extern FILE* stats_file;
+extern std::vector<json> core_stats;
 #endif
 
 class StatsCollectionScope final {
 public:
-  explicit StatsCollectionScope(const std::string &name) {
+  explicit StatsCollectionScope(const std::string &uri, const std::string& function, const std::string &operation_type) {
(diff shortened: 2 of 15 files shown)


src/include/utils/tiledb_helpers.h
```diff
@@ -0,0 +1,87 @@
...
+namespace detail {
+class StatsCollectionScope final {
+public:
+  explicit StatsCollectionScope(const std::string &name) {
+    tiledb::Stats::enable();
+    tiledb::Stats::reset();
+    name_ = name;
+  }
+
+  ~StatsCollectionScope() {
```
lums658, line 50: Is it possible to enable/disable stats through a config / environment variable?  It dumps alot of info and it would be nice to make that optional.  If not  we could turn it on or off via the command line and a (gasp) global variable.

src/include/feature_vector.h
```diff
@@ -58,14 +59,14 @@ using tdbFeatureVectorRange = tdbColMajorMatrix<T>
...
-    tiledb::Array array(ctx, uri, TILEDB_READ);
+    tiledb::Array array = tiledb_helpers::open_array(ctx, uri, TILEDB_READ);
     tiledb::Query query(ctx, array);
     tiledb::Subarray subarray(ctx, array);
     subarray.set_subarray(start, end);
     query.set_subarray(subarray);
     query.set_layout(TILEDB_ROW_MAJOR);
     query.set_data_buffer(attr_name, buffer);
-    query.submit();
+    tiledb_helpers::submit_query(query);
```
lums658, line 69: We should pass the URI and the function scope (there is a macro `tdb_func__` that can be used to get the function scope.  Right now query doesn't have enough context to be interpretable.

We might even consider passing `tdb_func__` to `open_array`.

src/include/utils/tiledb_helpers.h
```diff
@@ -0,0 +1,87 @@
...
+} // namespace detail
+
+/**
+ * @brief Opens a TileDB array and displays stats to stderr.
+ * 
+ * @param ctx The TileDB context to use.
+ * @param uri The URI of the array to open.
+ * @param query_type The mode to open the array.
+ */
+inline tiledb::Array open_array(const tiledb::Context &ctx,
```
lums658, line 68: These seem like useful wrappers in general.  There might be other things to do with them.

src/include/utils/tiledb_helpers.h
```diff
@@ -0,0 +1,87 @@
...
+class StatsCollectionScope final {
+public:
+  explicit StatsCollectionScope(const std::string &name) {
+    tiledb::Stats::enable();
+    tiledb::Stats::reset();
+    name_ = name;
+  }
+
+  ~StatsCollectionScope() {
+    fprintf(stderr, "---------- STATS %s ------------\n", name_.c_str());
```
lums658, line 51: Reporting stats should be fine-grained.

We should use the key-value logging singletons (dictionaries) I have set up in a recent `logging.h`.  Rather than dumping stats whenever an array is closed, you can stick the string into a dictionary instead.  Then, at the very end of the entire query, you can dump one or some or all of the stats via the dictionary.  This would be much cleaner and much more useful.

You could *even* use `std::regex` to dump just the information you want from a stats dump string.  Putting the info all in a dictionary that can be processed later gives you a centralized place to do this (or any) selective processing of stats.

This is how I have been managing logging information for gathering timing information and for gathering memory usage information.

src/include/utils/tiledb_helpers.h
```diff
@@ -0,0 +1,87 @@
...
+    tiledb::Stats::dump(stderr);
```
lums658, line 52: Where the output goes should be configurable using an `ostream` with a default to `stdout`.  (I recommend using `stdout` as the default `ostream` as it is easier to grep for things -- separately filtering `stderr` from `stdout` is really annoying.)

Again, by using a singleton dictionary, with deferred display of the stats messages, it is easy to choose where to send the output.  The processing of messages is done through the dictionary, not the stats collector.

lums658 (changes requested):
We should use singleton dictionaries as in recent versions of `logging.h` to give fine-grained and centralized control over what gets printed and when (and to what `ostream`).  This should be quite straightforward:
1. Create a new singleton dictionary in `logging.h` for stats.
2. Put the stats message into the dictionary rather than printing it.
3. Deal with the dictionary when, where, how you want to display the stats information.

teo-tsirpanis:
Thanks for the review, given that the main branch has significantly changed, I will have to start it over.

src/include/feature_vector.h
teo-tsirpanis: Done.

src/include/utils/tiledb_helpers.h
teo-tsirpanis: Done. Dumping stats is controlled with an `enable_stats` variable, that can be toggled with a Python API.

src/include/utils/tiledb_helpers.h
teo-tsirpanis: Done. The file stats are written to is configurable with a `stats_file` variable. For the Python API it defaults to stdout (and cannot change).

apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -14,6 +14,9 @@ using Ctx = tiledb::Context;
 bool global_debug = true;
 double global_time_of_interest;
 
+bool enable_stats = false;
+FILE* stats_file = stdout;
```
teo-tsirpanis, line 18: I tried commenting them out and did not get linker errors because the `extern` symbols were not found. Does anyone know why?

src/src/CMakeLists.txt
```diff
@@ -127,12 +127,15 @@ target_link_libraries(kmeans_lib INTERFACE
 #---[ Demos ]------------------------------------------------------------
 add_executable(flat_l2 flat_l2.cc)
 target_link_libraries(flat_l2 PUBLIC kmeans_lib nlohmann_json::nlohmann_json)
+target_compile_definitions(flat_l2 PUBLIC TILEDBVS_ENABLE_STATS)
```
lums658, line 130: Do we do the right thing somewhere in terms of downloading and installing `nlohmann_json`?  I had that in a local subdirectory (with the single-header `json.hpp`).  Do we want to just keeping using it that way for now?  I deleted it in my branch when I excised `json` but we can put it back.  My own preference would be to (eventually) be clean and use it as an external package, but that is not a high priority at all.

src/include/stats.h
```diff
@@ -53,7 +53,47 @@
 
 using json = nlohmann::json;
```
lums658, line 54: Hah!  I had previously been logging the timing and memory data to json but found dealing with that in pandas to be too annoying for my level of pandas expertise, so I went back to fixed-width columns and excised everything about json from my branch.  Will have to add it back but should not be a problem.  Perhaps I can learn enough pandas to use json for my own loging.  It would make some things much much simpler in terms of saving and dumping things and, provided I or someone has sufficient pandas expertise, to process.

lums658 (approved):
All looks good to me.  I'll have to kick the tires once a merge is done as I completely removed `json` support from the branch I am currently working on.

src/src/CMakeLists.txt
teo-tsirpanis: We keep a local copy of `json.hpp` [like the Core](https://github.com/TileDB-Inc/TileDB/blob/dev/external/include/nlohmann/json.hpp).

We can properly add it as a dependency at a later time (preferably with vcpkg).

PR #80, opened 2023-07-06
Fix segfault in finite ram queries

NikolaosPapailiou:


ihnorton (commented):
LGTM but @lums658 please double-check.

lums658 (approved):
C++ fixes are the correct ones to fix the issue that was identified.

PR #90, opened 2023-07-10
Distributed query implementation

NikolaosPapailiou:


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -170,3 +179,140 @@ def query(
             )
 
         return np.array(r)
+
+    def distributed_query(
```
teo-tsirpanis, line 183: How about we add a `mode` parameter to the existing `query` function? This way choosing between local and distributed execution will be more seamless.

apis/python/test/test_ingestion.py
```diff
@@ -154,22 +176,25 @@ def test_ivf_flat_ingestion_fvec(tmp_path):
...
             k=k,
-            nprobe=partitions,
+            nprobe=nprobe,
             use_nuv_implementation=True,
         )
     )
     assert accuracy(result, gt_i) > MINIMUM_ACCURACY
+
+    result = index_ram.distributed_query(
+        np.transpose(query_vectors), k=k, nprobe=nprobe, mode=Mode.LOCAL
```
ihnorton, line 198: Eventually we should harmonize the interfaces so that we don't need to transpose everywhere.

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -44,7 +53,7 @@ def query(
         nprobe: int = 1,
     ):
         """
-        Open a flat index
+        Query a flat index
```
teo-tsirpanis, line 56: Can we do distributed queries in plain flat indices?

ihnorton (approved):

apis/python/test/test_ingestion.py
NikolaosPapailiou: Done

apis/python/src/tiledb/vector_search/index.py
NikolaosPapailiou: Done

apis/python/src/tiledb/vector_search/index.py
NikolaosPapailiou: We can but we haven't implemented that yet

PR #93, opened 2023-07-12
Wrap vq_query_heap

ihnorton:
Adds bindings to vq_query_heap.

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -53,7 +53,7 @@ class FlatIndex(Index):
         nqueries: int = 10,
         nthreads: int = 8,
         nprobe: int = 1,
-        query_type="heap"
+        query_type="heap",
     ):
         """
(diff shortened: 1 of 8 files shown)


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -77,7 +79,15 @@ def query(
 
         targets_m = array_to_matrix(np.transpose(targets))
 
-        r = query_vq(self._db, targets_m, k, nqueries, nthreads)
+        if query_type == "heap":
+            if nqueries != 10:
+                logging.warning("nqueries is ignored for heap query type")
```
NikolaosPapailiou, line 84: Should we remove `nqueries` from the params?

NikolaosPapailiou (approved):

apis/python/src/tiledb/vector_search/index.py
ihnorton: Done, thanks

PR #96, opened 2023-07-18
Cut 10% of query time with this one weird trick.

lums658:
Remove `std::sqrt` from `L2` as sum of squares will have the same ordering as square root of the sum of squares.  When actual distances are required, we can take the square root then.  This seems to save around 10% of query time.


teo-tsirpanis:
Tests in `unit_gemm` need to be updated.

ihnorton (approved):

PR #107, opened 2023-07-31
Add pyproject dependency on scikit-learn

ihnorton:
It is a de-facto requirement due to use in ingestion. We might relax it to optional again if we have better import guards with a clear message.

### DIFF
--- a/apis/python/pyproject.toml
+++ b/apis/python/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 dependencies = [
     "tiledb-cloud>=0.10.5",
     "tiledb>=0.15.2",
-    "typing-extensions" # for tiledb-cloud indirect, x-ref https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/428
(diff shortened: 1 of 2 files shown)


Shelnutt2 (approved):

apis/python/pyproject.toml
```diff
@@ -21,11 +21,12 @@ classifiers = [
 dependencies = [
     "tiledb-cloud>=0.10.5",
     "tiledb>=0.15.2",
-    "typing-extensions" # for tiledb-cloud indirect
+    "typing-extensions" # for tiledb-cloud indirect, x-ref https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/428
```
Shelnutt2, line 24: Missing a `,` here

ihnorton, line 24:
```suggestion
    "typing-extensions", # for tiledb-cloud indirect, x-ref https://github.com/TileDB-Inc/TileDB-Cloud-Py/pull/428
```

PR #109, opened 2023-08-01
Set dtype automatically in IVFFlatIndex

ihnorton:


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -43,7 +43,7 @@ def ingest(
     size: int = 1
         Number of input vectors,
         if not provided use the full size of the input dataset
-    partitions: int = -1
+    partitions: int = 1
```
NikolaosPapailiou, line 46: -1 is used to auto-select the partitions. Using 1 will always use 1 partition. Is this what we want in this PR?

ihnorton: This was a symptom of a different problem, now fixed (and reverted to -1) in the next commit.

PR #118, opened 2023-08-02
Add storage format versions and store more metadata in group metadata

NikolaosPapailiou:
- Add storage format versions so we can keep backwards compatibility when changing storage names
- Write more metadata for the index in the Group metadata
- Remove `dtype` from the index APIs.
- Add a check for `nprobe<=parttions`
- Reformat the code using black

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -154,14 +154,14 @@ class IVFFlatIndex(Index):
 
         dtype = group.meta.get("dtype", None)
         if dtype is None:
-            schema = tiledb.ArraySchema.load(self.parts_db_uri)
+            schema = tiledb.ArraySchema.load(self.parts_db_uri, ctx=self.ctx)
             self.dtype = np.dtype(schema.attr("values").dtype)
         else:
             self.dtype = np.dtype(dtype)
 
         self.partitions = group.meta.get("partitions", -1)
         if self.partitions == -1:
-            schema = tiledb.ArraySchema.load(self.centroids_uri)
+            schema = tiledb.ArraySchema.load(self.centroids_uri, ctx=self.ctx)
             self.partitions = schema.domain.dim("cols").domain[1] + 1
 
     def query(
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -154,14 +154,14 @@ class IVFFlatIndex(Index):
 
         dtype = group.meta.get("dtype", None)
         if dtype is None:
-            schema = tiledb.ArraySchema.load(self.parts_db_uri, ctx=self.ctx)
+            schema = tiledb.ArraySchema.load(self.parts_db_uri, ctx=tiledb.Ctx(self.config))
             self.dtype = np.dtype(schema.attr("values").dtype)
         else:
             self.dtype = np.dtype(dtype)
 
         self.partitions = group.meta.get("partitions", -1)
         if self.partitions == -1:
-            schema = tiledb.ArraySchema.load(self.centroids_uri, ctx=self.ctx)
+            schema = tiledb.ArraySchema.load(self.centroids_uri, ctx=tiledb.Ctx(self.config))
             self.partitions = schema.domain.dim("cols").domain[1] + 1
 
     def query(
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -5,13 +5,9 @@ import logging
 
 import numpy as np
 from tiledb.vector_search.module import *
+from tiledb.vector_search.storage_formats import storage_formats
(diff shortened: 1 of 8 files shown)


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -134,31 +127,48 @@ def __init__(
...
             self._db = load_as_matrix(self.parts_db_uri, ctx=self.ctx, config=config)
             self._ids = read_vector_u64(self.ctx, self.ids_uri)
 
-        self._centroids = load_as_matrix(self.centroids_uri, ctx=self.ctx, config=config)
-
-        # TODO this should always be available
+        dtype = group.meta.get("dtype", None)
         if dtype is None:
-            self.dtype = self._centroids.dtype
+            schema = tiledb.ArraySchema.load(self.parts_db_uri)
```
Shelnutt2, line 157: Need to pass a tiledb.Ctx to this based on the config.

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -134,31 +127,48 @@ def __init__(
...
+            self.dtype = np.dtype(schema.attr("values").dtype)
         else:
-            self.dtype = dtype
-        self._index = read_vector_u64(self.ctx, self.index_uri)
+            self.dtype = np.dtype(dtype)
+
+        self.partitions = group.meta.get("partitions", -1)
+        if self.partitions == -1:
+            schema = tiledb.ArraySchema.load(self.centroids_uri)
```
Shelnutt2, line 164: Need to pass a tiledb.Ctx to this based on the config.

Shelnutt2 (changes requested):
Minor changes to ensure we pass user config to load array schema to honor any config settings (i.e authentication/access configs)

apis/python/src/tiledb/vector_search/index.py
NikolaosPapailiou: Done. 
Thanks!

apis/python/src/tiledb/vector_search/index.py
NikolaosPapailiou: Done. 
Thanks!

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -134,31 +127,48 @@ def __init__(
...
             self._db = load_as_matrix(self.parts_db_uri, ctx=self.ctx, config=config)
             self._ids = read_vector_u64(self.ctx, self.ids_uri)
 
-        self._centroids = load_as_matrix(self.centroids_uri, ctx=self.ctx, config=config)
-
-        # TODO this should always be available
+        dtype = group.meta.get("dtype", None)
         if dtype is None:
-            self.dtype = self._centroids.dtype
+            schema = tiledb.ArraySchema.load(self.parts_db_uri, ctx=self.ctx)
```
Shelnutt2, line 157: Can't use `self.ctx` because tiledb-py and tiledb-vector-search might be linked against different tiledb libraries. A context can't be shared across separate libraries and this will cause a segfault. If you are in conda or building locally and link against the same libtiledb then this does work, but in pypi and CI it won't. I made the same mistake yesterday.

Shelnutt2 (changes requested):

Shelnutt2 (approved):

apis/python/src/tiledb/vector_search/index.py
NikolaosPapailiou: That is something that we should be careful about in a consistent manner. Is there a way to make sure that both  tiledb-py and tiledb-vector-search are linked against the same tiledb lib?

apis/python/src/tiledb/vector_search/storage_formats.py
```diff
@@ -0,0 +1,18 @@
+storage_formats = {
+    "0.1": {
+        "CENTROIDS_ARRAY_NAME": "centroids.tdb",
+        "INDEX_ARRAY_NAME": "index.tdb",
+        "IDS_ARRAY_NAME": "ids.tdb",
+        "PARTS_ARRAY_NAME": "parts.tdb",
+        "PARTIAL_WRITE_ARRAY_DIR": "write_temp",
+    },
+    "0.2": {
```
ihnorton, line 9: It's a little confusing that we are adding this here but setting the default to 0.1

ihnorton: eg https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/118/files#diff-b0ba694c23a77c9684b7d183ae060fd615e208a54123ed3d384a3aaf9d30b214R51

ihnorton: nvm I see, that is intended as the "fallback" version where unspecified in the file -- which is for all existing files -- and we are bumping the format to use 0.2 names for all new ingestions. 👍

ihnorton (approved):

PR #121, opened 2023-08-03
Use ZstdFilter as default compression

NikolaosPapailiou:


apis/python/src/tiledb/vector_search/storage_formats.py
teo-tsirpanis, line 1: Why are filters dependent on the format version? If an old client tries to read a compressed array it would read it just fine, wouldn't it?

apis/python/src/tiledb/vector_search/storage_formats.py
```diff
@@ -1,17 +1,21 @@
...
         "PARTIAL_WRITE_ARRAY_DIR": "write_temp",
+        "DEFAULT_ATTR_FILTERS": None,
     },
     "0.2": {
         "CENTROIDS_ARRAY_NAME": "partition_centroids",
         "INDEX_ARRAY_NAME": "partition_indexes",
         "IDS_ARRAY_NAME": "shuffled_vector_ids",
         "PARTS_ARRAY_NAME": "shuffled_vectors",
         "PARTIAL_WRITE_ARRAY_DIR": "temp_data",
+        "DEFAULT_ATTR_FILTERS": tiledb.FilterList([tiledb.ZstdFilter(level=1)]),
```
Shelnutt2, line 18: Is there a reason we use level=1 and not leave the default (level 3)?

apis/python/src/tiledb/vector_search/storage_formats.py
NikolaosPapailiou: Yes this is just the default values used. It will not have any impact for previously ingested arrays

apis/python/src/tiledb/vector_search/storage_formats.py
NikolaosPapailiou: No this is just what I experimented with but I don't expect much difference for level 3. 
Changed this to use the default level

ihnorton (approved):

PR #127, opened 2023-08-11
Lums/sc 32145/return distances

lums658:
This draft PR updates the IVF queries to return top k distances in addition to top k neighbor indices.  

All of our search functions now use a fixed size min heap of to keep distances and indices.  The heap is templated on the type of distance and type of index, which are kept as a tuple in the heap.  The heap is sorted on the first element (the score).  To create a matrix of top k neighbors, the search functions sorted the heap and then copied the neighbor indices into a matrix.  

With the new API, queries can now be called as
```
auto&& [ D, I ] = query(...)
```

To update the query functions to return distances in addition to indices, this PR
* Created a new file scoring.h which includes most of the functionality previously in defs.h
* Deleted defs.h
* Created a function `consolidate` which consolidates distance and index information from a vector of vector of indices into a single vector (the 0th vector in the vector of vectors)
* Factored out the code to copy the index information in the heap structure to a matrix into new functions `get_top_k_from_heap` and two overloads of `get_top_k`
* Created augmented functions `get_top_k_from_heap_with_scores` and two overloads of `get_top_k_with_scores` that return a tuple of a distance matrix and an index matrix.
* Replaced the final logic in all of the ivf queries with `consolidate`  + `get_top_k_with_scores` (not all functions needed `consolidate`)
* Added extensive new tests in unit_fixed_min_heap.cc, unit_linalg.cc, and unit_scoring.cc, notably to validate the new `get_top_k` family of functions.
* Updated the ivf_flat C++ CLI program to use the new query API
* Updated the Python bindings in module.cc to use the new query API

In addition
* Propagated the BLAS macro to exclude BLAS code from the library so that the CLIs can be built without BLAS
* Added docstrings to the query functions in ivf/qv.h
* Added an initializer list constructor to `Matrix` to aid in creating tests


NikolaosPapailiou (commented):
Please make sure that this works correctly with the recent change of supporting external_ids https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/126/files

lums658:
> Please make sure that this works correctly with the recent change of supporting external_ids https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/126/files

Yikes, okay.  The version of the querying code that you modified in 126 is a bit behind what I branched from -- but it looks like external_ids are only used with flat queries?  That should be straightforward to address.

And to be clear -- do you want some aspects of the external_ids to go into 127, do you want me to merge 126 and 127, or do you just want to make sure that if you do the merge things will work?

NikolaosPapailiou:
> > Please make sure that this works correctly with the recent change of supporting external_ids https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/126/files
> 
> Actually I merged the latest main into 127 but need to address this particular issue. It is easy enough to add the external ids to the queries, but I am not sure how to make use of them when calling from the CLI program? I'll make overloads for now.
> 
> And just to be clear -- this is only for the flat queries, correct?
> 
> And second question -- are the external_ids just a permutation -- meaning the contents are 0 <= ids[i] < size(ids)?

This was only added to Flat queries because ivf queries already supported the functionality by using the ids array. 

In general, we don't want to assume that the object_ids are always the position of the vector in the input array. External IDs could be provided by the user and can be random numbers without any specific order or expectation for contiguous ranges. We might even want to extend this to different type of ids i.e. string values. This is especially important for supporting updates as users are going to add vectors with 'external_ids' that we can't generate out of a table index and delete 'external_ids' making the id space non contiguous.

PR #128, opened 2023-08-14
Add support for updates

NikolaosPapailiou:


src/include/detail/ivf/index.h
```diff
@@ -56,6 +57,7 @@ int ivf_index(
     tiledb::Context& ctx,
     const ColMajorMatrix<T>& db,
     const std::vector<ids_type>& external_ids,
+    const std::vector<ids_type>& deleted_ids,
```
lums658, line 60: I am not quite sure I understand this?  Shouldn't there be an updates array that includes additions and deletions, plus a validity array?  This was how it was described in https://www.notion.so/TileDB-Vector-search-support-for-updates-9ef808bcc5c24e7197403bc19bc342d1?pvs=4 (iiuc).

The interface I created for queries now takes an updates array, update ids, and a validity array.

src/include/detail/ivf/index.h
```diff
@@ -71,13 +73,15 @@ int ivf_index(
...
   {
-
     scoped_timer _{"shuffling data"};
+    std::unordered_set<ids_type> deleted_ids_set(deleted_ids.begin(), deleted_ids.end());
     std::vector<size_t> degrees(centroids.num_cols());
     std::vector<ids_type> indices(centroids.num_cols() + 1);
     for (size_t i = 0; i < db.num_cols(); ++i) {
-      auto j = parts[i];
-      ++degrees[j];
+      if (auto iter = deleted_ids_set.find(external_ids[i]); iter == deleted_ids_set.end()) {
```
lums658, line 81: I don't understand `==` here?  Is this creating an index of only the deleted elements?  What would done do with that?  How are additions plus deletes handled?

src/include/detail/ivf/index.h
```diff
@@ -109,16 +113,18 @@ int ivf_index(
     // be difficult to implement.  Even this algorithm is not trivial to
     // parallelize, because of the random access to the indices array.
     for (size_t i = 0; i < db.num_cols(); ++i) {
-      size_t bin = parts[i];
-      size_t ibin = indices[bin];
+      if (auto iter = deleted_ids_set.find(external_ids[i]); iter == deleted_ids_set.end()) {
```
lums658, line 116: Same question about `==`

lums658, line 116: This is going to be expensive.  (log(size(deleted_ids) * size(db)).  But I don't immediately see an obvious linear algorithm (or parallelization).  But if size(deleted_ids) << size(db) it might not be horrible.

src/include/detail/ivf/index.h
```diff
@@ -168,14 +175,15 @@ int ivf_index(
     external_ids = read_vector<ids_type>(ctx, external_ids_uri, start_pos, end_pos);
   }
   return ivf_index<T, ids_type, centroids_type>(
-      ctx, db, external_ids, centroids_uri, parts_uri, index_uri, id_uri, start_pos, end_pos, nthreads);
+      ctx, db, external_ids, deleted_ids, centroids_uri, parts_uri, index_uri, id_uri, start_pos, end_pos, nthreads);
 }
 
 template <typename T, class ids_type, class centroids_type>
```
lums658, line 181: We will need to refactor this into functions that deal with updates and those that don't.  We should also discuss index.h in general -- I didn't actually know it was being used.

src/include/detail/ivf/index.h
```diff
@@ -71,13 +73,15 @@ int ivf_index(
...
     for (size_t i = 0; i < db.num_cols(); ++i) {
-      auto j = parts[i];
-      ++degrees[j];
+      if (auto iter = deleted_ids_set.find(external_ids[i]); iter == deleted_ids_set.end()) {
+        auto j = parts[i];
+        ++degrees[j];
+      }
     }
     indices[0] = 0;
     std::inclusive_scan(begin(degrees), end(degrees), begin(indices) + 1);
```
lums658, line 87: For an updatable index we should probably be using a data structure that isn't contiguous.

src/include/detail/ivf/index.h
NikolaosPapailiou: Yes this is implemented on the Python side for now. The cpp ingestion just takes a list of `deleted_ids` to ignore from the input. This is mainly used during consolidation of updates and previous data. The previous data need to be filtered with this set of ids.

src/include/detail/ivf/index.h
NikolaosPapailiou: This is filtering the vectors that are deleted and should be skipped.

src/include/detail/ivf/index.h
NikolaosPapailiou: This is filtering the vectors that are deleted and should be skipped.

src/include/detail/ivf/index.h
NikolaosPapailiou: Avoided adding complexity to ingestions that don't need deleted_ids by only checking when `deleted_ids` is not empty

src/include/detail/ivf/index.h
NikolaosPapailiou: I am not sure I understand this comment. The degrees data structure is working as expected as we just filter out the deleted vectors.

src/include/detail/ivf/index.h
NikolaosPapailiou: Avoided adding complexity to ingestions that don't need `deleted_ids` by only checking the set when `deleted_ids` is not empty.

When we do need to check for `deleted_ids` we could try to have the `db` filtered once and then iterate over just the filtered vectors. I left this for a future PR. I am not sure if changing the loop order would give much performance benefit here but we can try to do a performance evaluation.

ihnorton (approved):

PR #129, opened 2023-08-18
Incorporate updates to queries

lums658:
This draft PR prototypes the approach and interface for adding updates (additions and deletions) to an ivf_flat query.  Since the development of the API was to intended to be iterative, only one function was initially prototyped: `qv_query_heap_infinite_ram`.
* The function takes three additional arguments: an updates array, update_ids, and a validity vector
* Overloads are provided with and without the original arguments
* A single function implements the two overloads using `constexpr` to include or omit the code for processing the updates.
* Per the Notion document, the algorithm performs the ivf query as normal (without updates). The updates are processed by first deleting elements from the top k information returned from the ivf query.  Then a flat query is conducted on the vectors to be added and those results are merged with the original results.
* The above process and API modifications can be readily applied to all other ivf query functions.

Since this is a prototype API, the functionality has not yet been tested.


src/include/detail/ivf/qv.h
```diff
@@ -244,10 +163,185 @@ auto qv_query_heap_infinite_ram(
...
+        deletes[delete_idx] = updates[i];
+        delete_ids[delete_idx] = update_ids[i];
+        ++delete_idx;
+      }
+      assert (insert_idx <= num_valid);
+      assert (delete_idx <= num_invalid);
+      assert (insert_idx + delete_idx == i + 1);
+    }
+
+    filter_top_k_scores(min_scores, delete_ids);
```
NikolaosPapailiou, line 196: The IVF result should be filter by all update_ids and not just deletion_ids. This is because there can be changes in vector values for already existing ids and in these cases we only need to consider the vector once in the flat query bellow.

src/include/detail/ivf/qv.h
```diff
@@ -188,14 +102,19 @@ auto qv_query_heap_infinite_ram(
...
     auto&& partitioned_db,
     auto&& centroids,
     auto&& q,
     auto&& indices,
     auto&& partitioned_ids,
     size_t nprobe,
     size_t k_nn,
+    auto&& updates,
+    auto&& update_ids,
+    auto&& update_validity,
```
NikolaosPapailiou, line 117: We can use a more intuitive set of arguments:
```
    auto&& updates,
    auto&& update_ids,
    auto&& delete_ids,
```

In this way we also don't need the code that splits updates into deletes and inserts. The filtering of the ivf results will need to happen using a union of delete_ids and update_ids. And the flat query can take all update vectors as input.

NikolaosPapailiou: To avoid copying data from Python we can also add some methods that read updates/deletes from an array uri. We can use the same schema as the one defined in https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/128/files#diff-b0ba694c23a77c9684b7d183ae060fd615e208a54123ed3d384a3aaf9d30b214R126
In this array schema, deletes are represented by empty vector attributes.

jparismorgan:
Closing for now, we can come back and reference this functionality as we need to.

PR #131, opened 2023-08-29
Draft C++ API

lums658:
This PR includes the first layer of the TileDB-Vector-Search C++ API.  The API is defined in terms of concepts, with the primary ones being `feature_vector` and `feature_vector_range`.  

As defined in concepts.h:
* A `feature_vector` is a contiguous `random_access_range` with the addition of `operator[]` on the container.  Moreover, the expression `dimension(v)` must be valid for a type to meet the requirements of `feature_vector`.
* A `feature_vector_range` is a container of `feature_vector`.  The ith `feature_vector` is obtained via `operator[]`.  The expressions `dimension(r)` and `num_vectors(r)` must be valid.  `feature_vector_range` is not actually a range (in the sense of `std::ranges`, so we will need to change the name at some point.  *Or* we could consider requiring `feature_vector_range` actually meet the requirements of `std::range` -- in which case it would need to have all the iterator plumbing, etc.

The PR also includes a set of customization point objects (CPOs) for the expressions `dimension` and `num_vectors` (implemented as niebloids).  The `dimension` CPO is defined so that, for instance, `std::vector<float>` is a `feature_vector` and `Matrix<float>` is a `feature_vector_range`.

The functionality provided by these two concepts is sufficient to implement our flat queries.  Additional concepts for `partitioned_feature_vector_range` (already started) will be needed for the IVF query functions.  The flat `qv_query_heap` function would now look like the following:

```c++
template <feature_vector_range DB, feature_vector_range Q, std::integral Index>
auto qv_query_heap(const DB& db, const Q& query, int k_nn, unsigned nthreads) {
  
  auto top_k = ColMajorMatrix<size_t>(k_nn, num_vectors(query));
  auto top_k_scores = ColMajorMatrix<float>(k_nn, num_vectors(query));

  auto min_scores = k_min_heap<float, Index>(k_nn);
    
  for (size_t j = 0; j < num_vectors(query); ++j) {
    feature_vector auto q_vec = query[j];
    for (size_t i = 0; i < num_vectors(db); ++i) {
      auto score = L2(q_vec, db[i]);
      min_scores.insert(score, i);
    }
    get_top_k_with_scores_from_heap(min_scores, top_k[j], top_k_scores[j]);
  }

  return std::make_tuple(std::move(top_k_scores), std::move(top_k));
}
```

Note that this completely eliminates any visibility of whether the inputs are row major or column major.

Using CPOs we can easily adapt other suitable types to meet the requirements of `feature_vector` and `feature_vector_range`.  For example (though I think this would be a bad idea, for many reasons, as listed in a prior Notion document) `std::vector<std::vector<>>`  We may need to use the `tag_invoke` mechanism rather than niebloids, however.

Unit tests are in unit_concepts.cc and unit_cpos.cc.


ihnorton:
#173+

PR #132, opened 2023-08-30
[draft] High-level Index API for dynamic language wrapping

ihnorton:
- Draft C++ Index API base class for simplified wrapping from high-level languages. This class (or something like it) is intended to wrap the strongly typed/type-specialized index API used for algorithm implementations, as outlined in #131. This class provides for a transition between NumPy arrays and the type-specialized containers required for each query function type. In particular, it is intended to remove the expensive/hard-to-maintain dispatching currently implemented in the pure Python layer.
- Sketch of private implementation pattern (hand-wavy) to hold the algorithm- and eltype-specialized index instances, with specialization based on vector element type of the stored index.


src/include/index.h
```diff
@@ -0,0 +1,90 @@
...
+
+class IndexOptions{};
+
+class TrainingParameters{};
+
+class FeatureVectorArray {
+  // .data() -> void*
+  // .ndim() -> size_t
+  // .size() -> size_t
+  // .extents() -> std::vector<size_t>
```
lums658, line 27: Is extents redundant with data() and ndim()?

src/include/index.h
```diff
@@ -0,0 +1,90 @@
...
+template <typename From, typename To>
+struct MapLike{};
+
+using StringMap = MapLike<std::string, std::string>;
+
+class IndexOptions{};
+
+class TrainingParameters{};
+
+class FeatureVectorArray {
```
lums658, line 23: This matches (modulo some naming) what is in api.h (in #131).

src/include/index.h
```diff
@@ -0,0 +1,90 @@
...
+    QueryVectorArray vectors,
+    size_t top_k,
+    QueryResults& results
+  ) -> std::pair(FeatureVectorArray, IdVector);
+
+  // Insert from in-memory array, optionally with IDs and training parameters
+  void insert(
+    FeatureVectorArray vectors,
+    std::optional<IdVector> ids = std::nullopt,
+    std::optional<TrainingParameters> params = std::nullopt
```
NikolaosPapailiou, line 73: `UpdateOptions`? Lets have a consistent naming with either `options` or `params`.
 Also it is not clear to me why we call this training params. In my mind here we can pass indexing parameters which are not only about training (e.g. threads, cpu, batch size etc.)

src/include/index.h
```diff
@@ -0,0 +1,90 @@
...
+
+  // Query this index with given vectors
+  auto query(
+    QueryVectorArray vectors,
+    size_t top_k,
+    QueryResults& results
+  ) -> std::pair(FeatureVectorArray, IdVector);
+
+  // Insert from in-memory array, optionally with IDs and training parameters
+  void insert(
```
NikolaosPapailiou, line 70: `insert` should also have URI based input to support batch ingestion

lums658:
I pushed a first implementation of the type-erased index class in #131.  The type-erased `Index` class is contained in `api.h` along with a type-erased `FeatureVector` class and a type-erased `FeatureVectorArray` class.  The `Index` class currently dispatches to a `flat_index` class, defined in `index.h`.  Constructors and a `query` function taking a `feature_vector_array` have been implemented (and compile correctly).

`FeatureVector` and `FeatureVector` are essentially complete and have been unit tested.

ihnorton:
Closing in favor of #210 (via #173 etc.)

PR #133, opened 2023-08-30
Add automatic update array fragment consolidation

NikolaosPapailiou:
Also fix query corner cases


ihnorton (approved):
Ok for follow-up if you prefer.

PR #134, opened 2023-09-08
Elide uses of ranges

lums658:
In some builds for Python on Mac, the compiler did not recognize `random_access_range`.  This PR temporarily removes all uses of `std::range`.


ihnorton (approved):
👍 thanks

teo-tsirpanis (approved):

NikolaosPapailiou (approved):

PR #141, opened 2023-09-14
Npapa/parallel update queries

NikolaosPapailiou:
Also fix some bugs with properly closing the updates array after each open operation.

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -1,3 +1,4 @@
+import multiprocessing as mp
 import numpy as np
 import sys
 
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -37,15 +39,32 @@ def __init__(
...
         if self.update_arrays_uri is None:
-            return internal_results_d[:, 0:k], internal_results_i[:, 0:k]
+            return self.query_internal(queries, k, **kwargs)
+
+        # Query with updates
+        # Perform the queries in parallel
+        retrieval_k = 2 * k
+        kwargs["nthreads"] = int(mp.cpu_count() / 2)
+        parent_conn, child_conn = mp.Pipe()
+        p = mp.Process(
```
ihnorton, line 51: Is there a reason not to use threading here? That will be a bit safer -- with with this, we'll need to call `mp.set_start_method('spawn')` to avoid issues with the default (`fork`) on linux.

ihnorton (approved):
LGTM, thanks for the update.

PR #144, opened 2023-09-19
Fix empty results

NikolaosPapailiou:
- Change `fixed_min_pair_heap` to update a heap even if it is not yet at max size
- Add deterministic (max uint, max float) fill values for empty results
- Initialize `min_scores` with correct `ids_type` from the template

### DIFF
--- a/src/include/detail/flat/vq.h
+++ b/src/include/detail/flat/vq.h
@@ -138,7 +138,7 @@ auto vq_query_heap(
   }
 
   consolidate_scores(scores);
-  auto top_k = get_top_k_with_scores(scores, k_nn);
+  auto top_k = get_top_k_with_scores<fixed_min_pair_heap<float, Index>, Index>(scores, k_nn);
 
   return top_k;
 }
@@ -223,7 +223,7 @@ auto vq_query_heap_tiled(
   } while (load(db));
 
   consolidate_scores(scores);
-  auto top_k = get_top_k_with_scores(scores, k_nn);
+  auto top_k = get_top_k_with_scores<fixed_min_pair_heap<float, Index>, Index>(scores, k_nn);
 
   return top_k;
 }
@@ -300,7 +300,7 @@ auto vq_query_heap_2(
   } while (load(db));
 
   consolidate_scores(scores);
-  auto top_k = get_top_k_with_scores(scores, k_nn);
+  auto top_k = get_top_k_with_scores<fixed_min_pair_heap<float, Index>, Index>(scores, k_nn);
 
   return top_k;
 }
--- a/src/include/detail/flat/vq.h
+++ b/src/include/detail/flat/vq.h
@@ -145,6 +145,9 @@ auto vq_query_heap(
   consolidate_scores(scores);
   auto top_k = get_top_k_with_scores(scores, k_nn);
 
+  debug_slice(std::get<0>(top_k));
+  debug_slice(std::get<1>(top_k));
+
   return top_k;
 }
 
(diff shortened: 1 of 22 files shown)


src/include/detail/flat/vq.h
```diff
@@ -223,7 +223,7 @@ auto vq_query_heap_tiled(
   } while (load(db));
 
   consolidate_scores(scores);
-  auto top_k = get_top_k_with_scores(scores, k_nn);
+  auto top_k = get_top_k_with_scores<fixed_min_pair_heap<float, Index>, Index>(scores, k_nn);
```
lums658, line 226: The `get_top_k` functions deduce the types from the arguments (or they do in my current branch).  I'll make sure all of those functions are consistent one way or the other and propagate those changes to this PR.

src/include/utils/fixed_min_heap.h
```diff
@@ -57,11 +57,8 @@ class fixed_min_set_heap_1 : public std::vector<T> {
 
   void insert(T const& x) {
     if (Base::size() < max_size) {
-      Base::push_back(x);
-      // std::push_heap(begin(*this), end(*this), std::less<T>());
-      if (Base::size() == max_size) {
-        std::make_heap(begin(*this), end(*this), std::less<T>());
-      }
+      this->push_back(x);
+      std::push_heap(begin(*this), end(*this), std::less<T>());
```
lums658, line 61: These all look good.  Previous was a case of over-optimization :-/.

src/include/scoring.h
```diff
@@ -306,6 +304,10 @@ inline void get_top_k_with_scores_from_heap(
       begin(min_scores), end(min_scores), begin(top_k), ([](auto&& e) {
         return std::get<1>(e);
       }));
+  for (size_t i = min_scores.size(); i < top_k.size(); ++i) {
```
lums658, line 307: I will make all of the `top_k` functions consistent and propagate to this PR.

lums658 (commented):
Most / all of these are present in the nn-descent/vamana branch.  I'll propagate those changes and also make idiom consistent across all of the `top_k` functions and all of the query functions.

lums658:
Updates to this PR:

- Propagated type parameterization in a consistent way through all query functions
  - Three types are generally gathered at the top of each query: a `feature_type`, an `id_type`, and a `score_type`.  The first two are obtained from arguments passed to the function.  `score_type` is set to be `float`.  All the type parameterized code in the search functions uses these types, so they can be changed by changing the types at the top of the function.
  - Eliminated some type parameters so that most functions can rely on type deduction and be called without explicit template type parameters.  The exceptions are the queries invoked with URIs, since the URIs stand in for the arguments from which we would be able to deduce types.
  - Updated module.cc to conform with propagated type parameterization.
- Added a `pad_with_sentinels` function to pad out any instances where the query results are smaller than the requested number of queries (generalized an propagated Nikos's fix).  Because of the way that scores are passed in, this works when `top_k` is a matrix and different scores might have different sizes.
- Added unit tests for `pad_with_sentinels`.
- Added unit_tests for the various `top_k` functions that exercise varying sizes of scores and top_k (i.e., that exercise the `pad_with_sentinels` being called by `top_k`).
- Got rid of `global_debug` and `global_verbose`.  We need a better way of having global control over printf debugging (perhaps via our logging infrastructure) but the globals were just getting to be too annoying and getting in the way of passing CI.
- Fixed a bug in `flat_l2`
- Verified all unit tests compile and run
- Verified `flat_l2` and `ivf_flat` produce correct answers for all available algorithms

PR #145, opened 2023-10-04
Fix missing flat ids

NikolaosPapailiou:
This was causing backwards compatibility issues for old arrays that don't have an ids array.

### DIFF
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -36,12 +36,15 @@ class FlatIndex(Index):
             ctx=self.ctx,
             config=config,
         )
-        self.ids_uri = self.group[
-            storage_formats[self.storage_version]["IDS_ARRAY_NAME"] + self.index_version
-        ].uri
-        if tiledb.array_exists(self.ids_uri, self.ctx):
-            self._ids = read_vector_u64(self.ctx, self.ids_uri, 0, 0)
-        else:
+        try:
+            self.ids_uri = self.group[
+                storage_formats[self.storage_version]["IDS_ARRAY_NAME"] + self.index_version
+            ].uri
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledb/vector_search/flat_index.py
```diff
@@ -36,12 +36,15 @@ def __init__(
             ctx=self.ctx,
             config=config,
         )
-        self.ids_uri = self.group[
-            storage_formats[self.storage_version]["IDS_ARRAY_NAME"] + self.index_version
-        ].uri
-        if tiledb.array_exists(self.ids_uri, self.ctx):
-            self._ids = read_vector_u64(self.ctx, self.ids_uri, 0, 0)
-        else:
+        try:
```
ihnorton, line 39: Do you mind to add some comment inline (so it doesn't have to be reconstructed from git-blame):
- about why we are doing this and what version/pre-version we need this for
- TODO that we should deprecate and remove support for the older version(s)

ihnorton (approved):
Minor request, otherwise LGTM.

apis/python/src/tiledb/vector_search/flat_index.py
Shelnutt2: Also is there a reason for the try block instead of checking for the group member?

NikolaosPapailiou: I couldn't find a method that checks for group member existence without throwing an exception. Is there such a method?

NikolaosPapailiou: Ok we can use standard contains operation for this. Applied

NikolaosPapailiou: Added a comment about the check and this being ok to not have ids provided. For backwards compatibility, I don't think we need to stop supporting previous versions atm.

PR #146, opened 2023-10-05
Add unit tests for cloud and demo notebooks

NikolaosPapailiou:
This adds:
- `TILEDB_REST_TOKEN` config for running tests. Local execution needs a valid token. For CI execution we pass the token from github secrets.
- Unit tests for cloud related functionality (distributed ingestion/query with `tiledb://` uris)
- Framework to unit test demo cloud notebooks
  - We use `nbmake` to run the notebooks as tests automatically


NikolaosPapailiou:
[sc-34621]

apis/python/test/common.py
```diff
@@ -183,3 +185,8 @@ def accuracy(result, gt, external_ids_offset=0, updated_ids=None):
         total += len(temp_result)
         found += len(np.intersect1d(temp_result, gt[i]))
     return found / total
+
+
+def random_name(name: str) -> str:
```
ihnorton, line 190: Please add a comment about why we use this.

ihnorton (approved):

apis/python/test/common.py
NikolaosPapailiou: Added

PR #148, opened 2023-10-09
Timetravel implementation

NikolaosPapailiou:
This is using the native TileDB time-traveling functionality to implement time-traveling for vector search.

Details:
- **Base arrays:** We only write to base arrays during batch ingestion. During ingestion we use a consistent timestamp to write all updates. We also keep track of all the ingestion timestamps in group metadata.
- **Update arrays:** Each time we write an update to the update array we can use different timestamps. For updates we keep the full history to allow arbitrary timetravel in past timestamps, even between ingestion timestamps.
- **Index open with timestamp:** Here we support both point in time and `[start, end]` time-traveling.
  - For point in time, we find the latest base array ingestion timestamp and use that for the base query. For the update query we use `[base_ingestion_timestamp, point_in_time]`
  - For `[start, end]` time-traveling: If `start > first_base_ingestion` we restrict the query to the update array using the same timestamp range. Otherwise, we also query the base array at the first_ingestion timestamp.


NikolaosPapailiou:
[sc-34496]

ihnorton (commented):
Please bump the storage version.

ihnorton (approved):

PR #149, opened 2023-10-14
Lums/sc 32552/nn descent

lums658:
This PR includes the implementation of two graph-based indexing algorithms: nn-descent and vamana.  The cpp_api branch was also merged into this one, so there is also code in this PR from that branch.

CPP API
======
For C++ library API
* api.h - Type-erased interface between Python and C++, including `Vector`, `FeatureVector`, and `Index`.  The type-erased classes take a URI and read in the corresponding TileDB array.  The underlying type of the `Vector`, `FeatureVector`, or `Index` class is determined by the data type of the TileDB array.
* concepts.h - Initial introduction of C++20 concepts.  A collection of concepts for defining the API to the vector search library.  This allows us to abstract away the format of the data storage (e.g., we don't need to worry about row-major or column-major).
* Use concepts for template parameters in function templates.  Replace type-specific calls in the library (e.g., in query functions, matrix classes) with CPOs.  There will be many small diffs due to this replacement.  The two most important concepts are

Further descriptions about the CPP_API PR can be found in PR 131.

Graph-Based Indexing and Search
=========================

* Graph classes
  * adj_list.h - Two adjacency list classes, one that just stores indices and one that stores indices and scores
  * nn-graph.h - An adjacency list class `nn_graph` that keeps the neighbors in a `fixed_min_heap` in order to store only the neighbors with the top-n scores

Graph creation and search algorithms
* nn-descent.h - Implements the nn-descent algorithm, using an `nn_graph`
* vamana.h - Implements a `vamana_index` class that builds an index according to the vamana algorithm as described in the "filtered fresh DiskANN" paper (*not* either the original DiskANN paper nor the fresh DiskANN paper).  
  * The index is built by iterating over the vectors in the data set and processing each vector as an insert into the index.  
  * When a vector is added, a best-first path from the medoid (the vector closest to the middle of the vector data set) is computed. 
  *  The path is then pruned to keep the degree of the vertexes on that path below a given threshold.  
  * The functions for performing the best-first search and the pruning are `greedy_search` and `robust_prune` (which are the names given in the original DiskANN paper).  

Utility and helper functions
* diskann.h - functions for reading DiskANN memory index files

Demo applications
* The `demo_*` programs in the test subdirectory are essentially integration tests that output qualitative data for assessing proper functioning of key algorithms in building the graph-based indexes (such as dumping graph information to disk so that the graphs can be read and plotted by networkx).  Two notebooks are provided for examining data dumped by `unit_vamana` and `vamana`: `plot_unit_vamana.ipynb` and `plot_vamana.ipynb`.

Unit tests
* Unit tests for the new classes and functions are provided in include/test.

CLI drivers
* Two CLI drivers are provided for vamana indexing: src/vamana/index.cc which builds a vamana-based index for a specified TileDB set of vector data, and src/vamana/query.cc that queries a vamana-based index.  As with previous C++ CLI programs, these print extensive logging data.

Benchmark scripts
* src/benchmarks/vamana - diskann.bash and vamana.bash respectively run indexing and querying programs from DiskANN and TileDB on some small-ish datasets (datasets that will fit in memory) and report timings (etc).  DiskANN is only configured to compile and run on x86 Linux, so these scripts need to be run on an appropriate platform.

Cleanup of unused files

*NB*  A number of the unit tests read and write arrays from the local file system.  These are hard-coded to my personal development environment since we are not currently pulling those into a uniform location (this should be fixed at some point so that test arrays are pulled in manually by running a script or automatically as part of cmake).


src/include/algorithm.h
```diff
@@ -186,6 +194,8 @@ void range_for_each(
            start,
            stop,
            f = std::forward<UnaryFunction>(f)]() mutable {
+            // scoped_timer __{tdb_func__ + std::string{" (lambda)"}, true};
```
Shelnutt2, line 197: Remove

src/include/concepts.h
```diff
@@ -36,84 +36,177 @@
...
+//   contiguous -- all the vectors are stored in a single contiguous array
+//   partitioned -- the range is a range of partitions, each of which is a
+//                  contiguous range of vectors
+//   distributed?-- the range is a range of partitions, each of which is a
+//
+// @todo operator()(size_t, size_t) ?
+// ----------------------------------------------------------------------------
+template <class D>
+concept feature_vector_array =
+    // feature_vector<inner_range_t<D>> &&
```
Shelnutt2, line 151: Remove?

src/include/concepts.h
```diff
@@ -36,84 +36,177 @@
...
+    // std::ranges::random_access_range<D> && /* std::ranges::sized_range<D> &&
+    // */ subscriptable_range<D> && requires(D d, const
+    // std::iter_difference_t<std::ranges::iterator_t<D>> n) {
+    requires(D d, size_t n) {
+  { num_vectors(d) } -> semi_integral;
+  { dimension(d) } -> semi_integral;
+  //{
+  //        d[n]
```
Shelnutt2, line 159: Remove?

src/include/detail/flat/qv.h
```diff
@@ -147,8 +142,10 @@ auto qv_query_heap(
     int k_nn,
     unsigned nthreads) {
   scoped_timer _{tdb_func__};
+  // load(db);
```
Shelnutt2, line 145: Remove?

src/include/detail/flat/qv.h
```diff
@@ -147,8 +142,10 @@ auto qv_query_heap(
...
 
-  // using feature_type = typename std::remove_reference_t<decltype(db)>::value_type;
+  // using feature_type = typename
+  // std::remove_reference_t<decltype(db)>::value_type;
```
Shelnutt2, line 148: Remove?

src/include/detail/flat/qv.h
```diff
@@ -226,30 +218,31 @@ auto qv_query_heap_tiled(
...
+  load(db);
 
-  // using feature_type = typename std::remove_reference_t<decltype(db)>::value_type;
-  using id_type = typename std::remove_reference_t<decltype(ids)>::value_type;
-  using score_type = float;
-
-  if constexpr (is_loadable_v<decltype(db)>) {
-    db.load();
-  }
+  // debug_slice(db);
```
Shelnutt2, line 223: Remove?

src/include/detail/flat/vq.h
```diff
@@ -184,25 +147,21 @@ auto vq_query_heap_tiled(
...
-    const ID& ids,
-    int k_nn,
-    unsigned nthreads) {
+    T, DB& db, const Q& q, const ID& ids, int k_nn, unsigned nthreads) {
   // @todo Need to get the total number of queries, not just the first block
   // @todo Use Matrix here rather than vector of vectors
 
-  // using feature_type = typename std::remove_reference_t<decltype(db)>::value_type;
+  // using feature_type = typename
+  // std::remove_reference_t<decltype(db)>::value_type;
```
Shelnutt2, line 155: Remove?

src/include/detail/flat/vq.h
```diff
@@ -245,46 +200,37 @@ auto vq_query_heap_tiled(
...
-    const Q& q,
-    const ID& ids,
-    int k_nn,
-    unsigned nthreads) {
+    T, DB& db, const Q& q, const ID& ids, int k_nn, unsigned nthreads) {
   // @todo Need to get the total number of queries, not just the first block
   // @todo Use Matrix here rather than vector of vectors
 
-  // using feature_type = typename std::remove_reference_t<decltype(db)>::value_type;
+  // using feature_type = typename
```
Shelnutt2, line 223: Remove?

Shelnutt2 (commented):
Minor cleanup comments, still reviewing the rest.

ihnorton:
#162, #173+

PR #150, opened 2023-10-18
Add APIs to create empty indexes

NikolaosPapailiou:
This introduces functionality to create empty vector indexes. 

It also adds support for using different partition sizes for different timestamps of the index history.

### DIFF
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -1,10 +1,12 @@
-import numpy as np
 import json
-
-from tiledb.vector_search.module import *
-from tiledb.vector_search.storage_formats import storage_formats, STORAGE_VERSION
-from tiledb.vector_search import index
 from typing import Any, Mapping
+
+import numpy as np
+
+from tiledb.vector_search import index
+from tiledb.vector_search.module import *
+from tiledb.vector_search.storage_formats import (STORAGE_VERSION,
+                                                  storage_formats)
 
 MAX_INT32 = np.iinfo(np.dtype("int32")).max
 TILE_SIZE_BYTES = 128000000  # 128MB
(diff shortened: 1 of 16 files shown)


apis/python/src/tiledb/vector_search/flat_index.py
```diff
@@ -80,10 +86,96 @@ def query_internal(
...
+            if "already exists" not in message:
+                raise err
+        group = tiledb.Group(uri, "w")
+        group.meta["dataset_type"] = index.DATASET_TYPE
+        group.meta["dtype"] = np.dtype(vector_type).name
+        group.meta["storage_version"] = STORAGE_VERSION
+        group.meta["index_type"] = INDEX_TYPE
+        group.meta["base_sizes"] = json.dumps([0])
+        group.meta["ingestion_timestamps"] = json.dumps([0])
+        tile_size = TILE_SIZE_BYTES / np.dtype(vector_type).itemsize / dimensions
```
ihnorton, line 121: We might want to move this to a function or class, along with the other calculation functions below, so that we can easily see where all of our "default size" computation happens.

apis/python/src/tiledb/vector_search/flat_index.py
```diff
@@ -80,10 +86,96 @@ def query_internal(
...
+
+        ids_array_name = storage_formats[STORAGE_VERSION]["IDS_ARRAY_NAME"]
+        parts_array_name = storage_formats[STORAGE_VERSION]["PARTS_ARRAY_NAME"]
+        ids_uri = f"{uri}/{ids_array_name}"
+        parts_uri = f"{uri}/{parts_array_name}"
+
+        if not tiledb.array_exists(ids_uri):
```
ihnorton, line 128: Should we error out if the array already exists, since this is a `create` function?

ihnorton (approved):
Minor comments, LGTM overall - looking forward to getting this in langchain!

apis/python/src/tiledb/vector_search/flat_index.py
NikolaosPapailiou: Applied

PR #155, opened 2023-11-16
Python bindings for type-erased FeatureVectorArray and IndexIVFFlat

lums658:
This PR builds on #154, adding Python bindings to the type-erased classes.  This is a straightforward wrapping of the classes, using pybind11.  The bindings are in `vector_search/module2.cc` and corresponding tests are in `test_module2.py`.  The file `test_module2.py` is a good source of examples of the interface.

An important feature of the binding for `FeatureVectorArray` is that it is non-oriented.  It can take a numpy array of either orientation (C or F).  A numpy array constructed from a `FeatureVectorArray` will have expected (row) orientation.  The assumption in all cases is that individual vectors are stored along the major axis of the array, i.e., the elements in each vector are stored contiguously.

The following snippets of code will all do the same thing.
```python
    a = np.array(np.random.rand(10000, 128), dtype=np.float32)
    b = vspy.FeatureVectorArray(a)
    c = np.array(b)

    a = np.array(np.random.rand(10000, 128), dtype=np.float32, order='F')
    b = vspy.FeatureVectorArray(a)
    c = np.array(b)

    a = np.array(np.random.rand(10000, 128), dtype=np.float32)
    b = vspy.FeatureVectorArray(a.T)
    c = np.array(b)
```

An example of using `IndexIVFFlat` to train and query would look like the following:
```python
       # Create the index and load training set and query set
       a = vspy.IndexIVFFlat(id_type="uint32", px_type="uint32")
       training_set = vspy.FeatureVectorArray(ctx, siftsmall_base_uri)
       query_set = vspy.FeatureVectorArray(ctx, siftsmall_query_uri)
 
        a.train(training_set, "random");
        a.add(training_set);

        s, t = a.query_infinite_ram(query_set, k_nn, nprobe)
```
A trained index can be saved to a TIleDB group.

Querying a trained index looks like the following:
```python
        a = vspy.IndexIVFFlat(ctx, siftsmall_flatIVF_index_uri_32_64)
        query_set = vspy.FeatureVectorArray(ctx, siftsmall_query_uri)

        s, t = a.query_finite_ram(query_set, k_nn, nprobe)
```

Testing
-------

```text
cd apis/python
pip3 install .
pytest test/test_module2.py
```

Legacy Bindings
----------------

The free-function query functionality prior to #153 is provided via the new interfaces in #153.  Indexing and query functionality is encapsulated in #153 in the `IndexIVFFlat` class, the `tdbPartitionedMatrix` class, and the `PartitionedMatrix` class, there were some ownership issues with arrays passed in from Python because the new classes want to own their data (and thus move any arrays passed directly to them).  An non-owning wrapper that bridges the legacy functionality to the new interface of the query functions is provided by a `PartitionedMatrixWrapper` class, that provides the same functionality as the `PartitionedMatrix` class, but is non-owning.

All of the tests in the Python test directory pass, except that I was not able to execute `test_cloud.py`.

Time Traveling
---------------

This branch was merged with `main` prior to the most recent push.  It thus includes the support for time-traveling that is currently in `main`.  The constructor for `IndexIVFFlat` now takes a timestamp, which is stored in the constructed `IndexIVFFlat` object.  Subsequent calls that read arrays are all done relative to that timestamp.

Calls for writing the index group are all made with the same timestamp, which is the time at which `write_index` is invoked (in milliseconds since the epoch).

Time traveling functionality has not been directly tested in C++ for this PR.


lums658:
Incorporated in #210.

PR #161, opened 2023-12-11
lums/sc 37627/indexivfflat Group schema

lums658:
This PR creates a group class `ivf_flat_index_group` for the `ivf_index` class that abstracts the `TileDB::Group` representing an IVF Flat Index.  It also rolls up preceding PRs 131, 149, 153, and 154.  The primary changes for this PR were to add the files `ivf_flat_group.h` and `ivf_flat_metadata.h` and to incorporate the revised access to group information into the `ivf_flat_index` class contained in `ivf_flat_index.h`.  Also added are unit tests for `ivf_flat_group` and `ivf_flat_metadata`.

The group and metadata classes to a fair amount of verification and testing of the data available in the TileDB::group and its associated metadata.

### Representing information about the `TileDB::Group`

The `TileDB::Group`  always contains the following arrays, which are the minimum necessary to represent an IVF Flat Index:
```text
  partition_centroids
  partition_indexes
  shuffled_vector_ids
  shuffled_vectors
```
Note that there is a level of indirection between `ivf_flat_index_group` and these names.  The names may change as TileDB-Vector-Search evolves.  A version number is used to distinguish different versions of the group data.

Other arrays that the Group may contain are:
```c++
  input_vectors
  external_ids
  updates
```

The Group contains another Group 
```c++
temp_data
```

Finally, the group contains information about attribute filters, and whether or not it supports time traveling, with the tags
```c++
default_attr_filters
support_timetravel
```
This information is captured by `ivf_flat_index_group`.  A `uniquie_ptr` to an `ivf_flat_index_group` is now part of the `ivf_flat_index` class.

### Index Arrays

The index represented by an `ivf_flat_index_group` support insertions, deletions, and consolidation.  As a result, the size of the matrix or vector being stored in one of the group arrays will change over time.  To handle this, the domain for each of the arrays is set to  go from zero to `std::numeric_limits<type>::max() - 1`.  To specify how much of each such array is active and should be read in, we keep information about the index dimension, its overall size (number of vectors and ids), and its number of partitions (equal to number of centroids and one less than the index size).  If time traveling is enabled, this information is updated on each consolidation.  The `ivf_flat_index_group` class has methods for getting the most recent value of each of these quantities, as well as to append a new value.  Access to arbitrary timestamps will be added as part of implementing time traveling.

### Metadata
The group also contains metadata with further information about the Group:
```c++
      dataset_type
      index_type
      storage_version
      dtype
      feature_type
      id_type
      indices_type
      base_sizes
[...]

PR #162, opened 2023-12-11
Merge

lums658:
This PR supersedes #149 -- it comprises `main` with a squash merge of #149.  The test below is from the original PR.

This PR includes the implementation of two graph-based indexing algorithms: nn-descent and vamana.  The cpp_api branch was also merged into this one, so there is also code in this PR from that branch.

CPP API
======
For C++ library API
* api.h - Type-erased interface between Python and C++, including `Vector`, `FeatureVector`, and `Index`.  The type-erased classes take a URI and read in the corresponding TileDB array.  The underlying type of the `Vector`, `FeatureVector`, or `Index` class is determined by the data type of the TileDB array.
* concepts.h - Initial introduction of C++20 concepts.  A collection of concepts for defining the API to the vector search library.  This allows us to abstract away the format of the data storage (e.g., we don't need to worry about row-major or column-major).
* Use concepts for template parameters in function templates.  Replace type-specific calls in the library (e.g., in query functions, matrix classes) with CPOs.  There will be many small diffs due to this replacement.  The two most important concepts are

Further descriptions about the CPP_API PR can be found in PR 131.

Graph-Based Indexing and Search
=========================

* Graph classes
  * adj_list.h - Two adjacency list classes, one that just stores indices and one that stores indices and scores
  * nn-graph.h - An adjacency list class `nn_graph` that keeps the neighbors in a `fixed_min_heap` in order to store only the neighbors with the top-n scores

Graph creation and search algorithms
* nn-descent.h - Implements the nn-descent algorithm, using an `nn_graph`
* vamana.h - Implements a `vamana_index` class that builds an index according to the vamana algorithm as described in the "filtered fresh DiskANN" paper (*not* either the original DiskANN paper nor the fresh DiskANN paper).  
  * The index is built by iterating over the vectors in the data set and processing each vector as an insert into the index.  
  * When a vector is added, a best-first path from the medoid (the vector closest to the middle of the vector data set) is computed. 
  *  The path is then pruned to keep the degree of the vertexes on that path below a given threshold.  
  * The functions for performing the best-first search and the pruning are `greedy_search` and `robust_prune` (which are the names given in the original DiskANN paper).  

Utility and helper functions
* diskann.h - functions for reading DiskANN memory index files

Demo applications
* The `demo_*` programs in the test subdirectory are essentially integration tests that output qualitative data for assessing proper functioning of key algorithms in building the graph-based indexes (such as dumping graph information to disk so that the graphs can be read and plotted by networkx).  Two notebooks are provided for examining data dumped by `unit_vamana` and `vamana`: `plot_unit_vamana.ipynb` and `plot_vamana.ipynb`.

Unit tests
* Unit tests for the new classes and functions are provided in include/test.

CLI drivers
* Two CLI drivers are provided for vamana indexing: src/vamana/index.cc which builds a vamana-based index for a specified TileDB set of vector data, and src/vamana/query.cc that queries a vamana-based index.  As with previous C++ CLI programs, these print extensive logging data.

Benchmark scripts
* src/benchmarks/vamana - diskann.bash and vamana.bash respectively run indexing and querying programs from DiskANN and TileDB on some small-ish datasets (datasets that will fit in memory) and report timings (etc).  DiskANN is only configured to compile and run on x86 Linux, so these scripts need to be run on an appropriate platform.

Cleanup of unused files

*NB*  A number of the unit tests read and write arrays from the local file system.  These are hard-coded to my personal development environment since we are not currently pulling those into a uniform location (this should be fixed at some point so that test arrays are pulled in manually by running a script or automatically as part of cmake).


jparismorgan:
Is this PR meant to go before #161, which looks like a superset of this? Generally seems good, though would be good to apply my review on #161 here before landing.

Approving, though it would be good to get another review here as well.

jparismorgan (approved):

lums658:
> Is this PR meant to go before #161, which looks like a superset of this? Generally seems good, though would be good to apply my review on #161 here before landing.
> 
> Approving, though it would be good to get another review here as well.

The idea here was to do a merge with the latest main so that there were no conflicts with main and so that it could be quickly approved and merged back into main.  From there, we can rebase #153, #154, #155, and #161 so that their diffs will not be so large and hopefully get them reviewed and approved relatively quickly so that we can make and keep main consistent with the latest goings on with C++.

apis/python/setup.py
```diff
@@ -34,6 +34,9 @@ def get_cmake_overrides():
     if val:
         conf.append("-DUSE_MKL_CBLAS={}".format(val))
 
+    conf.append("-DTileDB_DIR=/Users/lums/Contrib/dist")
```
NikolaosPapailiou, line 37: Remove

src/cmake/HighFive.cmake
NikolaosPapailiou, line 1: Why is this needed?

src/include/detail/graph/diskann.h
```diff
@@ -0,0 +1,137 @@
...
+#include <string>
+
+#include "detail/graph/adj_list.h"
+#include "scoring.h"
+
+auto read_diskann_data(const std::string& path) {
+  uint32_t npoints{0};
+  uint32_t ndim{0};
+
+  std::ifstream binary_file(path, std::ios::binary);
```
NikolaosPapailiou, line 47: This should be TileDB Array, what is the plan to use TileDB storage for diskann?

src/include/detail/graph/nn-descent.h
```diff
@@ -0,0 +1,470 @@
...
+    auto&& g, auto&& db, Distance distance = Distance()) {
+  size_t num_updates{0};
+  size_t nvertices{num_vertices(g)};
+  // g.copy_to_update_edges();
+
+  g.build_in_edges();
+
+  for (size_t i = 0; i < nvertices; ++i) {
+    for (auto&& [_, j] : out_edges(g, i)) {
+      // print_types(i, j);
```
NikolaosPapailiou, line 168: Remove commented lines here and in other parts of the file. If something need to remain add a comment for why it is needed.

src/include/detail/graph/nn-descent.h
```diff
@@ -0,0 +1,470 @@
...
+    for (auto&& j : in_edges(g, i)) {
+      num_updates += nn_descent_1_step(g, db, i, j);
+    }
+  }
+  // g.swap_all_update_edges();
+
+  return num_updates;
+}
+
+#if 0
```
NikolaosPapailiou, line 181: Remove dead code if not needed

src/include/detail/graph/nn-descent.h
```diff
@@ -0,0 +1,470 @@
...
+    }
+  }
+
+  return parents;
+}
+
+template <class T, std::integral ID, class Distance = sum_of_squares_distance>
+auto nn_descent_1_query(
+    const detail::graph::nn_graph<T, ID>& graph,
+    auto&& db,
```
NikolaosPapailiou, line 274: Having arguments specified as `auto` greatly reduces the readability of this code

src/include/detail/ivf/qv.h
```diff
@@ -1295,10 +1301,12 @@ auto nuv_query_heap_finite_ram_reg_blocked(
                 auto partno = p + part_offset;
 
                 auto quartno = partno;
-                if constexpr (!has_num_col_parts<decltype(partitioned_db)>) {
-                  quartno = active_partitions[partno];
-                }
-                (void)active_partitions;
+                //                if constexpr
+                //                (!has_num_col_parts<std::remove_cvref_t<decltype(partitioned_db)>>)
```
NikolaosPapailiou, line 1305: Please remove lines that are commented out. Same bellow

src/include/detail/linalg/matrix.h
```diff
@@ -1,6 +1,5 @@
-
 /**
- * @file   matrix.h
+ * @file   index.h
```
NikolaosPapailiou, line 2: nit matrix

src/include/detail/linalg/matrix.h
```diff
@@ -102,12 +169,15 @@ class Matrix : public stdx::mdspan<T, matrix_extents<I>, LayoutPolicy> {
     Base::operator=(Base{storage_.get(), num_rows_, num_cols_});
   }
 
+#if 0
```
NikolaosPapailiou, line 172: Avoid dead code

src/include/test/time_abstraction_penalty.cc
NikolaosPapailiou, line 1: Remove if not needed

src/include/test/unit_high_five.cc
NikolaosPapailiou, line 1: Remove

NikolaosPapailiou (commented):
This PR has a lot of features packaged in a single review cycle. 

In terms of release safety, we definitely need CI to be enabled and the test to pass. I would also prefer to split this in multiple PRs that contain the distinct feature functionalities (i.e. CPOs, concepts, Diskann, Vamana, C++ API)

ihnorton:
#173+

PR #164, opened 2023-12-12
Validate the `queries` vector shape

jparismorgan:
### What
Here we validate the shape of the vector(s) in `queries` to make sure it matches the input data, addressing https://app.shortcut.com/tiledb-inc/story/31507/vs-validate-query-vectors-shape.

### Testing
* Adds a unit test for both `FLAT` and `IVF_FLAT` which tests a query against `siftsmall_base.fvecs `
* Adds a unit test for both `FLAT` and `IVF_FLAT` which tests creating index's with varying numbers of columns in their dataset vectors, and then for each of these index's we query against it with varying numbers of columns in the query vectors. We do this for both one and two dimensional query vectors.
* Adds tests to confirm the query vector is one or two dimensional.

Tests pass except cloud b/c I haven't logged in yet (I thought I had but it seems something happened)
```
(TileDB-Vector-Search) ~/repo/TileDB-Vector-Search/apis/python pytest             ✹ ✭jparismorgan/query-vectors-shape 
================================================= test session starts =================================================
rootdir: /Users/parismorgan/repo/TileDB-Vector-Search/apis/python
collected 29 items                                                                                                    

test/test_api.py ....                                                                                           [ 13%]
test/test_cloud.py EE                                                                                           [ 20%]
test/test_index.py ......                                                                                       [ 41%]
test/test_ingestion.py ..............                                                                           [ 89%]
test/test_mod
```

### TODO
There is still a part of one test failing which I will look into as a follow-up. 

```
# TODO(paris): This will throw with the following error. Fix and re-enable, then remove 
# test_index_with_incorrect_num_of_query_columns_in_single_vector_query:
#   def array_to_matrix(array: np.ndarray):
#           if array.dtype == np.float32:
#   >           return pyarray_copyto_matrix_f32(array)
#   E           RuntimeError: Number of dimensions must be two
# Here we test with a query which is just a vector, i.e. [1, 2, 3].
# query = query[0]
# if num_columns_for_query == num_columns:
#     index.query(query, k=1)
# else:
#     with pytest.raises(TypeError):
#         index.query(query, k=1)
```

### DIFF
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -40,7 +40,7 @@ class FlatIndex(index.Index):
             + self.index_version
         ].uri
         schema = tiledb.ArraySchema.load(self.db_uri, ctx=tiledb.Ctx(self.config))
-        self.expected_query_dimensions = schema.shape[0]
+        self.dimensions = schema.shape[0]
         if self.base_size == -1:
             self.size = schema.domain.dim(1).domain[1] + 1
         else:
@@ -76,8 +76,8 @@ class FlatIndex(index.Index):
                     self.ctx, self.ids_uri, 0, self.size, self.base_array_timestamp
                 )
     
-    def get_expected_query_dimensions(self):
-        return self.expected_query_dimensions
+    def get_dimensions(self):
+        return self.dimensions
 
     def query_internal(
         self,
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -40,7 +40,7 @@ class FlatIndex(index.Index):
             + self.index_version
         ].uri
         schema = tiledb.ArraySchema.load(self.db_uri, ctx=tiledb.Ctx(self.config))
-        self.expected_query_columns = schema.shape[0]
+        self.expected_query_dimensions = schema.shape[0]
         if self.base_size == -1:
             self.size = schema.domain.dim(1).domain[1] + 1
         else:
@@ -76,8 +76,8 @@ class FlatIndex(index.Index):
                     self.ctx, self.ids_uri, 0, self.size, self.base_array_timestamp
                 )
     
-    def get_expected_query_columns(self):
-        return self.expected_query_columns
+    def get_expected_query_dimensions(self):
+        return self.expected_query_dimensions
 
     def query_internal(
         self,
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -40,6 +40,7 @@ class FlatIndex(index.Index):
             + self.index_version
         ].uri
         schema = tiledb.ArraySchema.load(self.db_uri, ctx=tiledb.Ctx(self.config))
+        self.expected_query_columns = schema.shape[0]
         if self.base_size == -1:
             self.size = schema.domain.dim(1).domain[1] + 1
         else:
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledb/vector_search/flat_index.py
```diff
@@ -40,6 +40,7 @@ def __init__(
             + self.index_version
         ].uri
         schema = tiledb.ArraySchema.load(self.db_uri, ctx=tiledb.Ctx(self.config))
+        self.expected_query_columns = schema.shape[0]
```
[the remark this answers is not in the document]
NikolaosPapailiou: No column major or row major is more about how we store data on disk. The client API shouldn't care about providing queries in the same order.

NikolaosPapailiou, line 43: nit: change the name to dimensions

apis/python/test/test_index.py
```diff
@@ -89,3 +92,101 @@ def test_ivf_flat_index(tmp_path):
 
     index = index.consolidate_updates()
     query_and_check(index, np.array([[2, 2, 2]], dtype=np.float32), 3, {0, 2, 4}, nprobe=partitions)
+
+def test_index_with_incorrect_dimensions(tmp_path):
+    indexes = [flat_index, ivf_flat_index]
+    for index_type in indexes:
+        uri = os.path.join(tmp_path, f"array_{index_type.__name__}")
+        index = index_type.create(uri=uri, dimensions=1, vector_type=np.dtype(np.uint8))
```
NikolaosPapailiou, line 100: Lets use more than one dimension e.g. 2 or 3 dimensions

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -126,6 +126,13 @@ def __init__(
         self.thread_executor = futures.ThreadPoolExecutor()
 
     def query(self, queries: np.ndarray, k, **kwargs):
+        if queries.ndim != 1 and queries.ndim != 2:
```
NikolaosPapailiou, line 129: I think that we currently just support 2 dimensional query matrix. This is why one of your tests is failing.

NikolaosPapailiou (commented):
Thanks for the extensive testing!

apis/python/src/tiledb/vector_search/index.py
jparismorgan: Hmm, we have some unit tests that query with a 1 dimensional query, i.e. here:
```
test_ingestion.py -> test_ivf_flat_ingestion_fvec()
    # Test single query vector handling
    _, result1 = index.query(query_vectors[10], k=k, nprobe=nprobe)
    assert accuracy(result1, np.array([gt_i[10]])) > MINIMUM_ACCURACY
```
Looking more into it, it seems that `IVF_FLAT` can support this, but not `FLAT`.

jparismorgan: Hmm, we have some unit tests that query with a 1 dimensional query, i.e. here:
```
test_ingestion.py -> test_ivf_flat_ingestion_fvec()
    # Test single query vector handling
    _, result1 = index.query(query_vectors[10], k=k, nprobe=nprobe)
    assert accuracy(result1, np.array([gt_i[10]])) > MINIMUM_ACCURACY
```
Looking more into it, it seems that `IVF_FLAT` can support this, but not `FLAT`.

Is this a scenario we want to support? If not then I can change this check and remove the test cases that use a 1 dimensional query. cc @lums658 / @ihnorton

apis/python/src/tiledb/vector_search/flat_index.py
jparismorgan: Sure, done.

NikolaosPapailiou: Thanks, please use just `self.dimensions` this variable might be important for testing in other parts of the code also.

NikolaosPapailiou (approved):

PR #167, opened 2023-12-13
Enforce `queries` has two dimensions

jparismorgan:
### What
Based on [this discussion](https://tiledb.slack.com/archives/C0537B4V7Q8/p1702471649730639), we'd like to enforce that `queries` has two dimensions (i.e. `[[1, 2, 3], [1, 2, 3], [1, 2, 3], etc]`).

### Testing
* Unit test for this case throws.
* Had to delete some old unit tests which test passing in a one dimensions `queries` vector.


jparismorgan:
Note that the CI is failing b/c if issues with Cloud, but I will wait to merge this until that is passing.

NikolaosPapailiou (approved):

PR #168, opened 2023-12-14
Test that we can index different storage versions and then query them

jparismorgan:
### What
Two changes:
1. Fix a bug where we couldn't create an index with storage version `0.1` and index against it. This PR originally was failing on `0.1`, but @NikolaosPapailiou helped point out the fix, so it's working now. See the history of this description to see the old failing error message.

2. Adding a test for storage version backward compatibility ([story](https://app.shortcut.com/tiledb-inc/story/34979/add-unit-tests-for-array-backward-compatibility)). Here we test that with the most up to date version of the engine we can index data in storage versions and then query that data. Note that this does not test if a user who indexed data previously will be able to query that data with the current version of the engine. To test that we'll need to periodically index data (i.e. with each release) and then have tests which query / consolidate / etc. that data.

### Testing
Tests pass:
```
(TileDB-Vector-Search) ~/repo/TileDB-Vector-Search/apis/python pytest test/test_ingestion.py -s
==================================================== test session starts =====================================================
rootdir: /Users/parismorgan/repo/TileDB-Vector-Search/apis/python
collected 15 items                                                                                                           

test/test_ingestion.py ..............Random initialization:
sklearn score: 6.619185447692871
tiledb score: 5.804458141326904
K-means++:
sklearn score: 3.9469563961029053
tiledb score: 4.8537516593933105
.

==================================================== 15 passed in 41.17s =====================================================
```


NikolaosPapailiou (commented):
Thanks for adding this! 

We were assuming that we have backwards compatibility. I would suggest to do a small investigation for version 0.1

NikolaosPapailiou:
Checking this, it seems that the problem is that we don't pass the `storage_version` parameter to `flat_index.create`, `ivf_flat_index.create` and `index.create_metadata` methods

jparismorgan:
> Checking this, it seems that the problem is that we don't pass the storage_version parameter to flat_index.create, ivf_flat_index.create and index.create_metadata methods

Good find thank you! I'll update that and add it to this PR. Changing to a draft while I get it working.

NikolaosPapailiou (approved):

PR #173, opened 2023-12-20
lums/sc 38296/time traveling support

lums658:
This PR is a continuation of #161, which added support for maintaining vector search indexes in TileDB groups.  A large number of diffs are carried along from previously un-merged commits.

#### Implementation

* For consistency, all APIs now use timestamps only.  Conversion to `TemporalPolicy` is done within functions that directly call TileDB functions.
* Refactoring of functions in tdb_io.h to enable array creation independently from writing to it.
* Change functions in tdb_io.h to use timestamps in addition to temporal policy (which are kept for now for backward compatibility).
* An `ivf_flat_index` is now created at a given timestamp.  If the timestamp is nonzero, its value is used for reading the index from a TileDB group or for writing (ingesting).  If the timestamp is zero for reading the most recent timestamp is used.  If the timestamp is zero for writing, the current time is used (ms since the epoch).
* All arrays within the group are written with the same timestamp.

#### Bug Fixes

* The fully general API for `tdbMatrix` was not implemented.  This was implemented and the code in tdb_matrix.h was significantly cleaned up and simplified.
* There was a counting error  in `tdbPartitionedMatrix` such that the largest possible number of partitions was not calculated properly.  This is needed to allocate sufficient space for the `PartitionedMatrix` when the `tdbPartitionedMatrix` is operating out of core.

#### Known Issues

* Several unit tests no longer pass: `unit_vamana` and `unit_pqflat_index` most notably.  These will be updated in PRs that will be upcoming shortly, and so these are marked `WILL_FAIL` in CMakeLists.txt.
* `ivf_index` queries no longer seem to be able to handle the case when the total number of vectors is smaller than `nlist`.  This is a very uncommon edge case and so the unit test that included this case has been commented out for the time being.
* We may need to use a variant for the timestamp argument to support single timestamps as well as pairs with the same API.
* Functionality for specifying timestamps has not yet been integrated into the C++ CLI programs.
* This implementation has not been rigorously compared with non-trivial indices.  Full testing will be as part of the insert/delete PR.


apis/python/src/tiledb/vector_search/ivf_flat_index.py
```diff
@@ -495,7 +507,7 @@ def create(
             centroids_array_rows_dim, centroids_array_cols_dim
         )
         centroids_attr = tiledb.Attr(
-            name="centroids",
+            name="values",
```
jparismorgan, line 510: Could this be the issue?

NikolaosPapailiou: We need to make sure that this is backwards compatible and the new code can read old assets

jparismorgan:
Closing for now, we can come back and reference this functionality as we need to.

PR #184, opened 2024-01-04
Add pypi package publishing in CI/CD

dudoslav:
This PR adds automatic pypi package publishing in CI/CD.

I still need to:

1. When to execute export
2. Set export repo to pypi not test.pypi

But first, I need to know if this approach is OK?


dudoslav:
https://test.pypi.org/project/tiledb-vector-search/ is the published version on test.pypi under my credentials

dudoslav:
The pipeline now finally passes with a small check:
https://github.com/dudoslav/TileDB-Vector-Search/actions/runs/7459145166

PR #185, opened 2024-01-06
Refactor group and metadata functionality to support different types of indexes

lums658:
This PR refactors the infrastructure for TileDB groups and metadata that are used to represent indexes.  It builds on, and supersedes, work started in `lums/sc-37627/indexivfflat-group-schema`, which was not turned into a PR.  Once #162 and #173 are merged, the changes in this PR should be fairly self-contained and reasonably-sized.

#### Features
* The refactoring includes a (shallow) class hierarchy that builds on the base `index_group` and `index_metadata` classes.
* We use CRTP to extend the `index_group` and `index_metadata` classes.  CRTP provides static polymorphism and also allows the base and derived classes to call each other's members.
* The URIs for arrays in a group are *relative* to the group URI (this differs from the Python).

#### Operational Overview
* A TileDB-Vector-Search index is represented by a TileDB group, which in turn comprises the arrays necessary for using the index.
* Any TileDB index class contains a group index class that provides the API for storing and retrieving data and metadata for the index.
* Index group classes for specific indexes are derived from `index_group` and the corresponding metadata classes are derived from `index_metadata`
    * The class for the flat l2 index is `flat_l2_index` defined in flat_l2_index.h.  It uses the base group and metadata classes, `index_group` and `index_metadata`, which are respectively defined in index_group.h and index_metadata.h
    * The class for the ivf flat index is `ivf_flat_index`, defined in ivf_flat_index.h.  It uses the group class `ivf_flat_index_ group`, defined in ivf_flat_group.h and `ivf_flat_index_metadata`, defined in ivf_flat_metadata.h.  `ivf_flat_index_group` and `ivf_flat_index_metadata` respectively derive from `index_group` and `index_metadata`.
* The functions that the "derived" classes need to provide to `index_group` are
```c++
append_valid_array_names_impl();
create_default_impl();
```
These are invoked by the base class for inserting the names of the arrays that will comprise the group, and for creating the corresponding TileDB arrays.
* The functions that the "derived" classes need to provide to `index_metadata` are
```c++
metadata_string_checks_impl()
metadata_arithmetic_checks_impl()
json_to_vector_impl()
vector_to_json_impl()
dump_json_impl()
```
These functions provide reading and writing of metadata for the associated derived group.
* To support updates to arrays in the index, all of the arrays in a group have an extent of MAXINT.  The actual size of the data stored in each array is kept as metadata within the group.

#### IVF Flat Index
* The arrays in the TileDB group for an IVF Flat index are
```
partition_centroids
partition_indexes
shuffled_vector_ids
shuffled_vectors
```

* The metadata maintained for the base Index are
```C++
// The first group are replicated from the Python API
base_sizes                       // A list of sizes
dataset_type                   // What this dataset is for ("vector_search")
dtype                                // A string representing the python type of the vector features
index_type                       // A string indicating what type of index this is ("IVFFlat")
ingestion_timestamps    // A list of timestamps at which data was originally ingested, or re-ingested from updates
storage_version              // What version of the index ("0.3")
temp_size                        // Size of the temporary storage

index_kind                         // An int corresponding to C++ enum for type of index (string is better in this case)
feature_datatype              // An int representing the TileDB datatype enum for the vector features
feature_type                      // A string representing the Python type of the vector features
id_datatype                        // TileDB datatype enum for vector ids
id_type                               // Python type of ids
```
Lists are kept in json arrays.

* Additional metadata maintained for the IVF Flat Index are:
```
partition_history               // List of partition sizes for each timestamp
px_datatype                       // TileDB datatype enum for partition indices
indices_type                      // Python type of partition indices
```
* Although the types of the features, ids, indices, and so forth can be read from the corresponding arrays, we keep that information as metadata so that we don't have to incur the overhead of opening an array just to check its type.
* There is some redundancy in terms of the information that is stored.  In a future PR we should streamline this.  There is danger in the use of the enum for `index_kind` -- it is not versioned and so could change if the enum definition is changed.

#### Notes
* A group representing an IVF Flat Index for the "siftsmall" example has been added to the repository.  The arrays within the group are not compatible with the non-updatable arrays used for testing of queries, etc. since the group arrays have infinite extent and the non-updatable arrays do not.  The default for opening a TileDB array representing a matrix or a vector is to use the extents as the matrix or vector dimensions.  Functions using the "infinite extent" arrays obtain the dimensions from the metadata and pass the dimensions explicitly to the matrix and vector readers.
* The PR currently only contains the group and metadata mappings for version 0.3, but infrastructure is in place to support other versions (older and newer)
* The PR should be compatible with the Python API for FlatL2 and IVFFlat indexes (version 0.3), though this has only been lightly tested
* PR #173 has been merged into this PR (and #162 was merged into #173).
* There are placeholders for flat pq and vamana indexes.  When those are implemented, the corresponding group and metadata classes will be created.  The derived functionality for the ivf flat index is not very large; we expect the classes for other indexes to be similarly straightforward.


lums658:
Superseded by #221.

PR #186, opened 2024-01-10
Update to TileDB 2.19.0

ihnorton:


NikolaosPapailiou (approved):

jparismorgan (approved):

dudoslav:
I am not sure if it is that much needed but don't we also need to bump up versions in `requirements-py.txt` and `pyproject.toml`? I have created a PR with that change in it already: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/188

ihnorton:
CLose in favor of #188

PR #193, opened 2024-01-12
Use relative paths for objects in groups

jparismorgan:
### What
Update how we store group members to use relative URIs instead of absolute URIs. 

Note that we use a helper function to do this because when we run this code through `test_cloud.py`, which has a index_uri like `tiledb://paris-morgan/s3://tiledb-paris/tiledb/groups/zzz_unittest_vector_search_IdNihhOwQt/test_flat_array` we fail with

```
/opt/homebrew/anaconda3/envs/TileDB-Vector-Search-2/lib/python3.9/site-packages/tiledb/vector_search/ingestion.py:2033: in ingest
    create_arrays(
/opt/homebrew/anaconda3/envs/TileDB-Vector-Search-2/lib/python3.9/site-packages/tiledb/vector_search/ingestion.py:422: in create_arrays
    flat_index.create(
/opt/homebrew/anaconda3/envs/TileDB-Vector-Search-2/lib/python3.9/site-packages/tiledb/vector_search/flat_index.py:203: in create
    group.close()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = test_flat_array GROUP


    def close(self):
        """
        Close a Group.
        """
>       self._close()
E       tiledb.cc.TileDBError: [TileDB::REST] Error: Error in libcurl PATCH operation: 
libcurl error message 'CURLE_OK';
HTTP code 500; 
server response data '{"code":901,"message":"failed to write member changes to core group: unable to determine array namespace/name from URI (expected \"tiledb://\u003cnamespace\u003e/\u003cname\u003e\"): shuffled_vector_ids","request_id":"ac565d16-2fb6-4a8b-9383-864948c7b343"}'.
```
This does not happen with `file:///private/var/folders/jb/5gq49wh97wn0j7hj6zfn9pzh0000gn/T/pytest-of-parismorgan/pytest-290/test_flat_ingestion_u80/array` when it's local.

Please do let me know if this is not correct and we should be able to use relative paths for tile

### Context
I found this bug when working on backwards compatibility testing. I had set up a [CI job](https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/7488465944/job/20382999090#step:6:212) which can create indices and check them in from the runner. But I found an issue where when I copy these indices to my local machine and try to load them I fail with:
```
FAILED test_backwards_compatability.py::test_query_old_indices - 
tiledb.cc.TileDBError: [TileDB::ArrayDirectory] Error: Cannot open array; Array does not exist.
```
This happens when loading `self.db_uri` because it is `file:///home/runner/work/TileDB-Vector-Search/TileDB-Vector-Search/backwards-compatability-data/data/0.0.0/flat_uint8/shuffled_vectors`, but on my computer it's stored under `/Users/parismorgan/repo/TileDB-Vector-Search-2/backwards-compatability-data/data/0.0.0/flat_uint8/shuffled_vectors`.
```
flat_index.py
     class FlatIndex(index.Index):
        ...
        self.db_uri = self.group[
            storage_formats[self.storage_version]["PARTS_ARRAY_NAME"]
            + self.index_version
        ].uri
```
When talking with Seth he let me know that `All group members should be stored with relative URIs not absolute`. So here we do that.

### Extra bug fix
Included in this is another small bug fix - in [this code](https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/193/files#diff-1f069806fd80be4565aeb1f2af024bd4363d02a250d8f377dd455e7a43218620L394) we have:
```
group.add(external_ids_array_uri, name=IDS_ARRAY_NAME)
```
But it should actually be:
```
group.add(external_ids_array_uri, name=EXTERNAL_IDS_ARRAY_NAME)
```
So we make this change as well.

### Testing
* Adds a helper `move_local_index_to_new_location` which will move the index to a new location on your computer and delete the old one.  Then update all tests in `test_ingestion.py` to move the index, load it, and perform some operations on it.
* Also confirmed that cloud tests still pass with this change.


apis/python/src/tiledb/vector_search/utils.py
```diff
@@ -4,6 +4,14 @@
 
 import tiledb
 
+def add_to_group(group, uri, name):
+    '''
+    Adds an object to a group. Automatically infers whether to use a relative path or absolute path.
+    '''
+    if 'tiledb://' in uri or 's3://' in uri:
```
NikolaosPapailiou, line 11: We should only do this for `tiledb://` uris

apis/python/test/common.py
```diff
@@ -310,3 +312,13 @@ def check_training_input_vectors(index_uri: str, expected_training_sample_size:
         assert training_input_vectors.shape[0] == expected_training_sample_size
         assert training_input_vectors.shape[1] == expected_dimensions
         assert not np.isnan(training_input_vectors).any()
+
+def move_local_index_to_new_location(index_uri):
```
ihnorton, line 316: (not blocking) I believe you could use a pytest fixture for this: https://docs.pytest.org/en/6.2.x/fixture.html

apis/python/src/tiledb/vector_search/utils.py
```diff
@@ -4,6 +4,14 @@
 
 import tiledb
 
+def add_to_group(group, uri, name):
+    '''
+    Adds an object to a group. Automatically infers whether to use a relative path or absolute path.
```
ihnorton, line 9: Please add a note about why and cross-ref the sc-<id> ticket number here so that we can trace back to be able to remove this eventually.

ihnorton (approved):

apis/python/test/common.py
jparismorgan: Hmm yeah, I think you could. Though I don't totally see what benefit you get here, as this is a function we need to call rather than a variable we want to make available to all tests?

apis/python/src/tiledb/vector_search/utils.py
jparismorgan: Thanks, done.

apis/python/src/tiledb/vector_search/utils.py
NikolaosPapailiou: Lets add a TODO and. an issue to capture this and try to fix in the future.

NikolaosPapailiou (approved):

PR #195, opened 2024-01-15
Add `.clang-format` file and CI job to check for required formatting

jparismorgan:
### What
* Adds a `.clang-format` file.
* Adds a CI job to check that the code was formatted and fail if not.

This code is all pulled from this job with some minor updates (like upgrading to clang-format 17, renaming scripts to all follow a consistent pattern, and adding more comments): https://github.com/TileDB-Inc/TileDB/blob/dev/.github/workflows/check-formatting.yml.

Next up will be doing the same for python code using the pattern from: https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/.github/workflows/tiledb-cloud-py.yaml#L15.

### Testing
* Here you can see a job from the [fix run_clang_format.sh](https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/195/commits/5d3ec5ae953d99c0b687a1401e492d99cd91a3ae) commit where we fail because we need to make formatting changes: https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/7527895087/job/20488976085
* And here you can see the current job where we pass the job because all formatting is up to date: https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/7528038294/job/20489406766?pr=195


documentation/Building.md
```diff
@@ -66,4 +66,19 @@ docker build -t tiledb/tiledb-vector-search .
 You run the example docker image which provides the python package with:
 ```
 docker run --rm tiledb/tiledb-vector-search
+```
+
+# Formatting
+
+The C++ code is formatted with `clang-format` version 17. Install it yourself or by running:
```
ihnorton, line 73: Can we get 16? Core is using 16 so that will be easier to keep aligned between projects.

jparismorgan: So it's simple on linux. But on Mac `brew` doesn't have 16: https://formulae.brew.sh/formula/clang-format - it only has 17:
```
(TileDB-Vector-Search-4) ~/repo/TileDB-Vector-Search-4 brew install clang-format@16
Warning: No available formula with the name "clang-format@16". Did you mean clang-format, clang-format@11 or clang-format@8?
==> Searching for similarly named formulae and casks...
==> Formulae
clang-format                   clang-format@11                clang-format@8

To install clang-format, run:
  brew install clang-format
```
Looking more at Core, I think they actually have a bug, as they do:
```
install_apt_pkg() {
  wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
  add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-16 main' &&
  apt-get update -qq && apt-get install -qq -y clang-format-16
}

install_brew_pkg() {
  brew upgrade && brew install clang-format
}
```
But this will install 16 on linux and 17 on mac. I'm guessing that many people just ran `brew install clang-format` before 17 was out, but anyone new will have 17?

NikolaosPapailiou (approved):
Thanks for adding this!

documentation/Building.md
ihnorton: I'm ok with (1) then.

ihnorton (approved):

PR #197, opened 2024-01-16
Consolidate test data between C++ and Python into single folder

jparismorgan:
### What
Consolidate test data between C++ and Python into single folder. It looks like the C++ code wasn't actually using this data before as the tests pass with no changes, but this way they can use this data going forward.

### Testing
CI passes.


apis/python/test/common.py
```diff
@@ -7,6 +7,14 @@
 import tiledb
 from tiledb.vector_search.storage_formats import storage_formats, STORAGE_VERSION
 
+def get_test_data_path():
+    current_dir = os.path.dirname(__file__)
+    test_data_path = os.path.join(current_dir, "../../../test-data")
```
lums658, line 12: We need to have an absolute path for this.  The Clion debugger, for instance, will run the tests from a different directory.  One can get the top-level python directory (similar to CMAKE_SOURCE_DIR).

lums658, line 12: The location from previous PRs has been (at the top level):
```
TileDB-Vector-Search/external/test_data
├── README.md
├── arrays
│   ├── bigann10k
│   ├── fmnistsmall
│   ├── siftsmall
│   └── siftsmall_uint8
└── files
    ├── diskann
    └── siftsmall
```
For broader testing there are arrays with different feature types, different dimensions, and different numbers of vectors.  Attached is the directory tree at the next level down.  
[test_data_tree.txt](https://github.com/TileDB-Inc/TileDB-Vector-Search/files/13953355/test_data_tree.txt)

Per some other discussions, we should probably add a manual mechanism for also including the full sift and full bigann1M arrays for testing and/or benchmarking.  That can be a manual download.  There is also (iirc) a preprocessor macro to enable that in src/include/test/array_defs.h.

lums658, line 12: Speaking of src/include/test/array_defs.h -- I don't know if there is an easy way to do this, but it would be useful to automatically derive the array names and array paths from a single location for both Python and C++ (i.e., automatically generate array_defs.h and the array names and paths used by Python from our spec).

Maybe an aspirational story.

apis/python/test/test_index.py
```diff
@@ -13,6 +13,10 @@ def query_and_check(index, queries, k, expected, **kwargs):
         result_d, result_i = index.query(queries, k=k, **kwargs)
         assert expected.issubset(set(result_i[0]))
 
+@pytest.fixture(scope="module")
+def test_data_path():
```
lums658, line 17: See comment above on absolute vs relative path as well as naming.

apis/python/test/test_index.py
```diff
@@ -112,9 +116,9 @@ def test_index_with_incorrect_dimensions(tmp_path):
         # Okay otherwise.
         index.query(np.array([[1, 1, 1]], dtype=np.float32), k=3)
 
-def test_index_with_incorrect_num_of_query_columns_simple(tmp_path):
-    siftsmall_uri = "test/data/siftsmall/siftsmall_base.fvecs"
-    queries_uri = "test/data/siftsmall/siftsmall_query.fvecs"
+def test_index_with_incorrect_num_of_query_columns_simple(tmp_path, test_data_path):
+    siftsmall_uri = f"{test_data_path}/siftsmall/siftsmall_base.fvecs"
```
lums658, line 120: For these we should use a variable.  In PR #173 (sorry for overloading that with so many changes) to make sure the testing is consistent wrt to file naming I created a file test/array_paths.py that defined variables for most of the test arrays and files:
[array_paths.txt](https://github.com/TileDB-Inc/TileDB-Vector-Search/files/13953198/array_paths.txt)

We should use that convention (both in terms of using variables rather than hard-coded names, as well as putting those into test/array_defs.py).

There are many duplications of this below -- this comment applies to each of them (I won't comment on each of them).

test-data/README.md
```diff
@@ -0,0 +1,5 @@
+# Test Data
```
lums658, line 1: Again, should be  
[test_data_tree.txt](https://github.com/TileDB-Inc/TileDB-Vector-Search/files/13953357/test_data_tree.txt)

(Applies to all changes below.)

I can send you a tar file with all of this directory if it would be helpful.

lums658 (changes requested):
Main things that need to be addressed:

1. Use a root for the test that that is an absolute instead of relative path
2. Use variable names for the complete path to each array / file -- all such variable names should be kept in test/array_defs.py 
[array_paths.txt](https://github.com/TileDB-Inc/TileDB-Vector-Search/files/13953382/array_paths.txt)

3. Use the paths / files / naming convention I included in 
[test_data_tree.txt](https://github.com/TileDB-Inc/TileDB-Vector-Search/files/13953381/test_data_tree.txt)

(These are all in #173.)

apis/python/test/common.py
jparismorgan: Mmm I'm a bit confused... specific questions:
1. Are you using Clion to run the Python tests? If so, could you share with me your setup?
2. Are you saying that CLion will copy the `test_foo.py` and `common.py` files into a different temporary directory and then run them? And that's why relative path won't work?
3. I don't see `array_defs.py` in #173, I just see `array_defs.h`? Could you send me a link to `array_defs.py`?

Screenshot 2024-01-16 at 6 14 59 PM

lums658: Not 100% sure of how Clion does this, tbh — but, yes, I think it copies.  When running things under the debugger there is definitely an original and a copied version of the .py files. Also, CLion doesn’t run pytest directly, when invoking unit tests.  I did some poking around in the toolchain but didn’t see anything obvious.  Sufficiently advanced technology, unfortunately.

Mea culpa, I typed too fast — in Python I called the fille array_paths.py.  We should keep the same name between Python and C++ — I don’t have a strong preference between array_defs and array_paths.

I believe I attached array_paths.py to my earlier comments (had to rename it array_paths.txt) to my comments (github helpfully turns it into a hyperlink).  I'll send the .py file in slack.

Best Regards,
Andrew Lumsdaine

jparismorgan: Hmm I tested in CLion (running in both `Run` and `Debug` mode) and the relative paths look okay to me:

Screenshot 2024-01-16 at 9 40 42 PM

Screenshot 2024-01-16 at 9 13 38 PM

Is there some other configuration I should be using to see this issue?

NikolaosPapailiou (approved):

jparismorgan:
Closing in favor of other PRs. But for future reference, relative paths are required for the tests to work across different machines and with the repo saved to different paths on computers.

PR #198, opened 2024-01-16
Remove the `pip install git+github...` instructions

nickvigilante:
Spoke with John about this, and he echoes those instructions are confusing.


jparismorgan (approved):
Thanks, I agree, I was also a bit confused this (thoughI don't have _too_ much python experience).

PR #200, opened 2024-01-17
Add `pre-commit` to format files

jparismorgan:
### What
Here we add `pre-commit` to format files in the repo. We use the same configuration as https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/.pre-commit-config.yaml, but we also add `clang-format`.

To keep this easier to review, I'll split the code into three PRs:
1. This PR to add `pre-commit`.
2. A PR to run `pre-commit` on all files.
3. A PR to add a CI job which runs `pre-commit` (and runs it again in case any new files made it in which require formatting - this is still nicer to review than combining them).

Note that I did ignore some things in `ignore = ["F403", "F405", "E501", "E741"]` - these are mainly for line length, undefined imports, and other things which don't get automatically fixed. It would be nice to not ignore these, but ignoring them helps get this in faster and keep churn down. I would propose in a later follow-up we can add these back if we'd like.

### Testing
* #196 runs a CI job succesfully.
* I have checked that manually running `clang-format` vs running `pre-commit` result in the same formatting changes.
* I have run it locally and can get everything formatted well:
```
(TileDB-Vector-Search) ~/repo/TileDB-Vector-Search pre-commit run --all-files
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...............................................................Passed
Check for added large files..............................................Passed
prettier.................................................................Passed
ruff.....................................................................Passed
black....................................................................Passed
clang-format.............................................................Passed
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,31 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+      - id: check-yaml
+      - id: check-added-large-files
+
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.7.1
+    hooks:
+      - id: prettier
+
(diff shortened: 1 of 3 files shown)


NikolaosPapailiou (approved):
Thanks for this! 
Lets try to avoid ignoring some of the failures in a following PR

PR #204, opened 2024-01-19
Incorporate graph search code into main

lums658:
Graph-Based Indexing and Search
=========================

This is essentially a recap of one half of #149, omitting CLI drivers, demo applications, and benchmark scripts for compactness.  

* Graph classes
  * adj_list.h - Two adjacency list classes, one that just stores indices and one that stores indices and scores
  * nn-graph.h - An adjacency list class `nn_graph` that keeps the neighbors in a `fixed_min_heap` in order to store only the neighbors with the top-n scores

Graph creation and search algorithms
* nn-descent.h - Implements the nn-descent algorithm, using an `nn_graph`
* vamana.h - Implements a `vamana_index` class that builds an index according to the vamana algorithm as described in the "filtered fresh DiskANN" paper (*not* either the original DiskANN paper nor the fresh DiskANN paper).  
  * The index is built by iterating over the vectors in the data set and processing each vector as an insert into the index.  
  * When a vector is added, a best-first path from the medoid (the vector closest to the middle of the vector data set) is computed. 
  *  The path is then pruned to keep the degree of the vertexes on that path below a given threshold.  
  * The functions for performing the best-first search and the pruning are `greedy_search` and `robust_prune` (which are the names given in the original DiskANN paper).  

Utility and helper functions
* diskann.h - functions for reading DiskANN memory index files

Test data in `external/test_data`
* `siftsmall` test arrays
* `fmnistsmall` test arrays
* `diskann` test files

PR #205, opened 2024-01-19
Add concepts and cpos

lums658:
Concepts and CPOS
===============

This is a recap of part of #149.

Concepts and customization point objects:
* concepts.h - Initial introduction of C++20 concepts.  A collection of concepts for defining the API to the vector search library.  This allows us to abstract away the format of the data storage (e.g., we don't need to worry about row-major or column-major).  The two most important concepts are

Next steps (this PR blocks the following two):
* Use concepts for template parameters in function templates.  Replace type-specific calls in the library (e.g., in query functions, matrix classes) with CPOs.  This will be cross-cutting and there will be many small diffs due to this replacement.
* Create type-erased `FeatureVector` and `FeatureVectorArray` classes.


src/include/concepts.h
```diff
@@ -36,78 +36,196 @@
...
+    // */ subscriptable_range<D> && requires(D d, const
+    // std::iter_difference_t<std::ranges::iterator_t<D>> n) {
+    requires(D d, size_t n) {
+      //      typename D::feature_type;
+      { num_vectors(d) } -> semi_integral;
+      { dimension(d) } -> semi_integral;
+      //{
+      //        d[n]
+      //      } ->
+      //      std::same_as<std::iter_reference_t<std::ranges::iterator_t<D>>>;
```
jparismorgan, line 169: (Minor) Let's remove this commented out code?

src/include/concepts.h
```diff
@@ -36,78 +36,196 @@
...
+      { d[n] } -> feature_vector;  // Maybe redundant
```
jparismorgan, line 170: Did you determine whether it's redundant or not? Would be good to either remove this or add more context to the comment, as a future reader (i.e. me 😅) will be quite confused about this.

src/include/concepts.h
```diff
@@ -36,78 +36,196 @@
...
+// partitioned_feature_vector_range concept (WIP)
+// ----------------------------------------------------------------------------
+
+template <class D>
+concept partitioned_feature_vector_array =
+    feature_vector_array<D> && partitionable<D>;
+//    && requires(D d) {
+//      typename D::id_type;
+//      typename D::indices_type;
+//    };
```
jparismorgan, line 202: (Minor) Remove, please. Will stop reviewing for this, but let's remove all commented out code please.

src/include/cpos.h
```diff
@@ -0,0 +1,485 @@
...
+    return t.load();
+  }
+
+  template <class T>
+    requires(!_member_load<T>)
+  auto constexpr operator()(T&& t) const noexcept {
+    // return t.load();
+    return false;
+  }
+};
```
jparismorgan, line 306: What's going on here? Should we remove `// return t.load();` because we're just using `requires(!_member_load<T>)`?

src/include/test/unit_concepts.cc
```diff
@@ -27,132 +27,366 @@
...
+  CHECK(!partitionable<decltype(d)>);
+  CHECK(!partitionable<std::span<int>>);
+
+  CHECK(!partitionable<Vector<int>>);
+}
+
+template <class T>
+class dummy_feature_vector : public std::vector<T> {
+ public:
+  using base = typename std::vector<T>;  //::vector??
```
jparismorgan, line 328: (Minor) What is this comment saying?

src/include/test/unit_concepts.cc
```diff
@@ -27,132 +27,366 @@
...
+TEST_CASE("concepts: partitioned_feature_vector_range", "[concepts]") {
 }
 
-TEST_CASE("vector_database_0", "[concepts]") {
-  vector_database_0<float> db;
-  CHECK(b(db));
-  // REQUIRE(feature_vector<feature_vector_0<float>>);
+TEST_CASE(
+    "concepts: contiguous_partitioned_feature_vector_range", "[concepts]") {
 }
```
jparismorgan, line 392: Should these be removed?

src/include/test/unit_concepts_vs.cc
```diff
@@ -0,0 +1,177 @@
...
+  CHECK(dimensionable<Vector<bool>>);
+
+  CHECK(vectorable<Vector<int>>);
+  CHECK(vectorable<Vector<double>>);
+  CHECK(vectorable<Vector<bool>>);
+
+  CHECK(feature_vector<Vector<int>>);
+  CHECK(feature_vector<Vector<double>>);
+  CHECK(feature_vector<Vector<bool>>);  // @todo ? Vector of bool is not
+                                        // contiguous?
```
jparismorgan, line 78: (Minor) Could you add some more context here about what the todo is here (and everywhere below)? As it reads, I'm not sure if it's a todo or not, which is a bit confusing.

jparismorgan (commented):
Looking quite nice, mainly minor issues.

NikolaosPapailiou (commented):
The code in this PR looks good to me. 
Please fix the build and tests.

src/include/detail/graph/nn-descent.h
```diff
@@ -0,0 +1,470 @@
...
+    for (auto&& j : in_edges(g, i)) {
+      num_updates += nn_descent_0_step(g, db, i, j);
+    }
+  }
+  // g.swap_all_update_edges();
+
+  return num_updates;
+}
+
+// Local join
```
NikolaosPapailiou, line 119: Can you add some more comments on the difference of nn_descent_0 and nn_descent_1? Also picking some more descriptive name would improve the readability.

src/include/test/unit_vamana.cc
```diff
@@ -0,0 +1,1119 @@
...
+  CHECK(size(top_k) == 15);
+  CHECK(size(top_k_scores) == 15);
+  CHECK(size(visited) == 15);
+
+  for (size_t i = 0; i < size(top_k); ++i) {
+    std::cout << "( " << top_k[i] << ", " << top_k_scores[i] << " ), ";
+  }
+  std::cout << std::endl;
+
+#if 0
```
NikolaosPapailiou, line 261: Cleanup if not needed

src/include/utils/fixed_min_heap.h
NikolaosPapailiou, line 1: Can you explain a bit what the difference is in this file?

NikolaosPapailiou (approved):

src/include/concepts.h
lums658: I have not determined that.  Determining this will take some experimentations, which we should defer since it is not harmful at any rate.

src/include/test/unit_concepts_vs.cc
lums658: Added a comment:  "There is an open question about whether we should consider Vector of bool bo be a contiguous range or not.   It is not contiguous in the sense that it is not a contiguous array of bools, but it is contiguous in the sense that  it is a contiguous array of bits.  For now, if it passes the concept check, we'll let it be."

src/include/detail/graph/nn-descent.h
lums658: I think those both will be deprecated in a future PR.

ihnorton (approved):

ihnorton:
(whoever merges this, please use **squash merge**)

src/include/detail/graph/diskann.h
```diff
@@ -0,0 +1,144 @@
...
+  for (size_t node = 0; node < num_nodes; ++node) {
+    uint32_t num_neighbors;
+    binary_file.read((char*)&num_neighbors, 4);
+    for (size_t i = 0; i < num_neighbors; ++i) {
+      uint32_t id;
+      binary_file.read((char*)&id, 4);
+      g.add_edge(node, id, sum_of_squares(x[node], x[id]));
+    }
+    // @todo ??? Is this right ???
+    binary_file.seekg(max_degree - num_neighbors, std::ios_base::cur);
```
jparismorgan, line 137: Looks like the changes you made on the other PR didn't make it over here: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/204/files#diff-66f0c6afcad02cfba5854fbdfd2241b0068368e87818d1b3cfc8e14428f99c62

src/include/detail/graph/nn-descent.h
```diff
@@ -0,0 +1,470 @@
...
+ */
+
+template <
+    std::integral I,
+    std::integral J,
+    class Distance = sum_of_squares_distance>
+auto nn_descent_0_step(
+    auto&& g, auto&& db, I i, J j, Distance distance = Distance()) {
+  size_t num_updates{0};
+  for (auto&& [_, k] : out_edges(g, j)) {  // _ is dist(j, k)
```
jparismorgan, line 74: Looks like the commit I made to cleanup the code in the other PR did not make it over to this one: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/204/commits/6a1d1f69ce9949bb6d9936752de819fa13a033d2

src/include/detail/graph/vamana.h
```diff
@@ -0,0 +1,1018 @@
...
+   * @return True if the feature vectors are the same, false otherwise
+   */
+  bool compare_feature_vectors(const vamana_index& rhs) {
+    for (size_t i = 0; i < ::num_vectors(feature_vectors_); ++i) {
+      for (size_t j = 0; j < ::dimension(feature_vectors_); ++j) {
+        auto lhs_val = feature_vectors_(j, i);
+        auto rhs_val = rhs.feature_vectors_(j, i);
+        if (lhs_val != rhs_val) {
+          std::cout << "lhs_val != rhs_val" << std::endl;
+          // return false;
```
jparismorgan, line 995: Should this be commented back in?

jparismorgan (approved):
Still some issues with bringing code over from the #204 to this PR, and there are also unresolved comments on #204. But fine by me to land to keep moving forward.

PR #207, opened 2024-01-22
Enable PyPI upload in Build Wheels workflow

dudoslav:
This PR aims to enable upload to PyPI automatically when the `cibuildwheel` passes (another PR that enables tests is ready).

The desired behaviour:

1. Build all wheels
2. * Build sdist - currently not available until https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/181 will get merged
3. Upload wheels to test.pypi.org (this has to change to original)
4. Download and tests uploaded wheels (this is redundant after tests are enabled)

My suggestions:

1. I would reccomend to run cibuildwheel step on every commit in default branch, not only release commits. This way we would be fairly confident that the release workflow will pass. The upload step would only run on release event.


.github/workflows/build-wheels.yml
```diff
@@ -62,14 +62,19 @@ jobs:
...
     # needs: generate_backwards_compatibility_data
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.buildplat[0] }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, macos-12, windows-2022]
-        # `windows-2022, ` blocked by https://github.com/pybind/pybind11/issues/3445#issuecomment-1525500927
+        buildplat:
+          - [ ubuntu-22.04, manylinux_x86_64 ]
+#          - [ ubuntu-22.04, musllinux_x86_64 ]
```
ihnorton, line 73: remove commented?

ihnorton:
> I would recommend to run cibuildwheel step on every commit in default branch, not only release commits. This way we would be fairly confident that the release workflow will pass.

If we do the full build matrix that is a lot of extra jobs; I would lean toward only running the full wheel build check as part of nightlies.

PR #210, opened 2024-01-23
Type erased feature vector and feature vector array classes

lums658:
## Type Erased FeatureVector and FeatureVectorArray
This is a recap of the feature vector and feature vector array type-erasure component of #154.  

* The primary contribution of this PR is the implementation of type-erased `FeatureVector` and `FeatureVectorArray`.  Since one of the goals of having the type-erased classes is to seamlessly integrate with Python, the Python bindings for these classes were also added.
* Some updating of matrix and vector I/O was done since the type-erased classes were just copied over.  To update matrix and vector, the files implementing them were copied over the existing files, with some minimal changes.
* The cpo `num_vectors()` replaced `size()` in numerous places.

### Files Added (Copied from #173)
* include/api/README.md
* include/api/api_defs.h
* include/api/feature_vector.h
* include/api/feature_vector_array.h
* include/tdb_defs.h
* include/test/test_utils.h
* include/test/unit_api_feature_vector.cc
* include/test/unit_api_feature_vector_array.cc
* external/test_data/arrays/bigann10k

### Files Copied Over Previous
* include/detail/ivf/vq.h
* include/detail/linalg/linalg_defs.h
* include/detail/linalg/matrix.h
* include/detail/linalg/tdb_io.h
* include/detail/linalg/vector.h

### Files Modified
In addition, the following had small modifications (mostly to change `size()` to `num_vectors()`):
* include/algorithm.h
* include/detail/flat/qv.h
* include/detail/flat/vq.h
* include/detail/ivf/dist_qv.h
* include/detail/ivf/index.h
* include/detail/ivf/partition.h
* include/detail/ivf/qv.h
* include/detail/linalg/tdb_matrix.h
* include/detail/linalg/tdb_partitioned_matrix.h
* nclude/scoring.h
* include/test/CMakeLists.txt
* include/test/unit_ivf_qv.cc
* include/test/unit_ivf_vq.cc
* src/ivf_flat.cc

### Added (Copied from #173)
* python/src/tiledb/vector_search/module2.cc
* python/test/test_module2.py

### Modified
* python/CMakeLists.txt
* python/src/tiledb/vector_search/module.cc
* python/src/tiledb/vector_search/__init__.py
* python/CMakeLists.txt

**NOTE:**  The type-erased Python binding files that were copied over included code for index classes.  This code has been temporarily commented out (usually with #ifdef 0), pending the next PR.

## Arrays
* bigann10k (this is resulting in a large number of reported changed files)

## Overview of Type Erasure

(See also the README.md in include/api).

Type erasure is accomplished as a three-layer cake:
* A non-templated abstract base class with pure virtual functions for the member functions that we want to expose from the typed C++ classes.
* An implementation class that inherits from the abstract base class.  It is templated on the concrete class to be wrapped and keeps a member variable that is an object of the class to be wrapped.  Its member functions (all of which are overrides of the abstract base class member functions) forward to the underlying C++ implementation by invoking the appropriate member.  At this point, they internal data that is stored by the typed member variable also needs to be converted to the appropriate type before invoking the member variable function.
* A non-templated class that presents the user API.  It internally defines the abstract base class and the implementation class.  It has a `std::unique_ptr` to the abstract base class as a member variable.  During construction (either by passing in already constructed vectors or by reading the index from a TileDB group, the appropriate template types for the internal data to be stored by the internal implementation are inferred and an object of the implementation class is constructed and stored in the `std::unique_ptr`.

To illustrate the basic idea, consider `FeatureVector`.  In abbreviated form, where we just show a single function 'data', looks like this:
```c++
class FeatureVector {
    FeatureVector(const tiledb::Context& ctx, const std::string uri) {
      // get type of vector stored in uri array -- say, float
      feature_vector_ = std::make_unique<vector_impl<tdbVector<float>>>(ctx, uri);
    }
    auto data() {
     return feature_vector_->data();
   }

   class Base {
    virtual void* data() = 0;
  };

  template <class T>
  class Impl {
     explicit Impl(T&& t)
        : impl_vector_(std::forward<T>(t)) {
    }
    T impl_vector_;
  };

  std::unique<Base> feature_vector_;
};

The constructor to read the `FeatureVector` from a TileDB array has the following prototype:
```c++
FeatureVector(const tiledb::Context& ctx, const std::string& uri);
```
When that constructor is invoked, it first reads the schema associated with the `uri` and creates an implementation object based on that type.  For example, if the type read from the schema (`feature_type`) is one of a `float` or `uint8`, the constructor dispatches like this:
```c++
switch (feature_type_) {
      case TILEDB_FLOAT32:
        vector_ = std::make_unique<vector_impl<tdbVector<float>>>(ctx, uri);
        break;
      case TILEDB_UINT8:
        vector_ = std::make_unique<vector_impl<tdbVector<uint8_t>>>(ctx, uri);
        break;
}
```
At this point, we have created a `std::unique_ptr` of the abstract base class that points to an object of the derived class.
If we invoke the `data` member function of the outer (type-erased) `FeatureVector` class, we dispatch to the corresponding member of the object stored in the `std::unique_ptr`:
```c++
 auto data() const {
    return feature_vector_->data();
  }
```
Since `feature_vector_` actually points to the derived implementation class, its `data` member function is then invoked:
```c++
void* data() override {
      return impl_vector_->data();
    }
```
We return a `void*` since `data()` is an override of the non-templated `Base` class.
(TODO: In a future PR maybe we can cast to an appropriate type extracted from the type of `vector_`?)
(TODO: Is there a way to condense the boilerplate that is currently contained in all of these?)


src/include/detail/linalg/matrix.h
NikolaosPapailiou, line 1: nit: there are a lot of #if definitions in this file. Please cleanup the ones that are not needed.

src/include/detail/linalg/tdb_io.h
```diff
@@ -173,24 +279,30 @@ void write_vector(
...
+ *
+ * @todo Return a Vector instead of a std::vector
  */
 template <class T>
 std::vector<T> read_vector(
     const tiledb::Context& ctx,
     const std::string& uri,
     size_t start_pos,
     size_t end_pos,
-    const tiledb::TemporalPolicy temporal_policy = {}) {
```
NikolaosPapailiou, line 188: Why do we change the `TemporalPolicy` arguments in this PR? If we prefer `size_t` arguments for some reason, a `TemporalPolicy` should be replaced by 2 `size_t` timestamps to be able to cover both cases of `Timetravel` and `TimestampStartEnd`. This is especially important for read methods that need to implement `TimestampStartEnd` type reads.

ihnorton: https://app.shortcut.com/tiledb-inc/story/39936/re-expose-temporalpolicy-and-wrap-in-python

apis/python/src/tiledb/vector_search/module2.cc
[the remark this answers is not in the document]
lums658: Yes, that is a much better name.  "There are only two hard things in Computer Science: cache invalidation and naming things."

src/include/detail/linalg/matrix.h
lums658: Done.

src/include/detail/linalg/tdb_io.h
lums658: The current PR is equivalent to state of #173, in which Python only invokes functions from a single C++ file: index.h.
The function `ivf_index()` is the only one called from Python and the read of the `tdbColMajorMatrix` is the only place where something other than a `tiledb::TimeTravel` policy is used.  A `TemporalPolicy` interface to `tdbMatrix` is in place for this PR, and the call for creating the centroids matrix uses the `TimeStampStartEnd` policy.  So nothing differs functionally from any earlier PRs that had `TemporalPolicy` more widely used.  Properly propagating `TemporalPolicy` throughout is deferred for another PR.

In general the API needs better encapsulation of temporal [start, stop) as well as better encapsulation of spatial [start, stop).  See my comments in Story 39936, linked above.

PR #215, opened 2024-01-24
Partitioned Matrix API

lums658:
### Partitioned Matrix
 This PR moves the code for the refactored `PartitionedMatrix` from #173.  It recapitulates part of #153, which is primarily targeted at the `ivf_index` class. 
* The refactored `PartitionedMatrix` is a separate class and is used as a base class for `tdbPartitionedMatrix`. The `PartitionedMatrix` class encapsulates the partitioning indices, the partitioned vectors, and the partitioned ids and they are always self-consistent with each other and are completely contained in memory.  
* All of the management of out-of-core indices and so forth is provided by `tdbPartitionedMatrix`.
* This refactoring results in a significantly cleaner API for the query code.  The prior API included separate parameters for partitioned vectors, etc., while the new API uses the encapsulation of the `PartitionedMatrix`.
* The refactoring results in significantly simpler implementations (and may allow unification of finite and infinite RAM cases -- though that should be a later PR).
* Moved the updated Python bindings that use the new API for the query functions.
* Moved the test data siftsmall flat files from #173.

The `PartitionedMatrix` class and the refined API are important for implementing the `ivf_index` class.


apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -95,66 +133,73 @@ static void declare_pyarray_to_matrix(py::module& m, const std::string& suffix)
...
-         size_t nthreads) -> py::tuple { //std::pair<ColMajorMatrix<float>, ColMajorMatrix<size_t>> { // TODO change return type
+         size_t nthreads) -> py::tuple {
+
+        auto mat = ColMajorPartitionedMatrixWrapper<T, Id_Type, Id_Type>(
+            parts, ids, indices);
+
+        auto top_centroids = detail::ivf::ivf_top_centroids(
+            centroids, query_vectors, nprobe, nthreads);
+        auto r = detail::ivf::qv_query_heap_infinite_ram(
+            top_centroids, mat, query_vectors, nprobe, k_nn, nthreads);
```
NikolaosPapailiou, line 176: what is the purpose of this change?

apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -209,250 +269,419 @@ static void declare_nuv_query_heap_finite_ram(py::module& m, const std::string&
...
+
+            for (size_t i = 0; i < num_rows; ++i) {
+              py::list sublist = py::cast<py::list>(ptr[i]);
+              size_t sublist_length = py::len(sublist);
+              active_queries.emplace_back();
+              active_queries.back().reserve(sublist_length);
+              for (size_t j = 0; j < sublist_length; ++j) {
+                active_queries.back().emplace_back(py::cast<indices_type>(sublist[j]));
+              }
+            }
```
NikolaosPapailiou, line 526: Why are we changing the distributed query implementation in this PR. This could break our distributed execution if the new libraries are not present in the UDF images in cloud.

apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -209,250 +269,419 @@ static void declare_nuv_query_heap_finite_ram(py::module& m, const std::string&
...
+      part_uri,
+      active_partitions,
+      query,
+      active_queries,
+      indices,
+      id_uri,
+      k_nn,
+      timestamp);
+}, py::keep_alive<1,2>());
+#if 0
```
NikolaosPapailiou, line 539: Please avoid `#if 0` definitions. IMO they greatly reduce the readability of the code.

NikolaosPapailiou (commented):
This PR contains large changes in the Python module especially for distributed query execution. Can you please explain why they are required in this PR?

Changes in distributed execution libraries need to be released in coordination with cloud UDF images in order to avoid breaking production.

apis/python/src/tiledb/vector_search/module.cc
lums658: The API for all of the query functions (`qv`, `nuv`, etc.) changed to take relevant centroids, the partitioned matrix (encapsulating partitioned vectors, partitioned ids, and the partitioning index), and the queries.  So the calls to the C++ query functions from Python required some setup to call via the new API.  This is also intended to be transitional since we will (presumably) be implementing the indexes in C++ at some point, in which case there will also be a more sensible Python API.

apis/python/src/tiledb/vector_search/module.cc
lums658: This is an additional API, not a different one that takes a `py::array` rather than `std::vector<std::vector<int>>` for the `active_queries`.  The previous function is available starting at line #604 (obscured by the #if 0, I suppose).

apis/python/src/tiledb/vector_search/module.cc
lums658: Was keeping those for easy reference to previous implementation.  Will remove.

NikolaosPapailiou (approved):

PR #216, opened 2024-01-25
ivf_index class

lums658:
This PR recapitulates the `ivf_index` class portion of PR #153.

### `class ivf_index`

Specifically for `ivf_index`, the PR includes
* Completion of the `ivf_index` class, including implementation of `train()`, `add()`, and multiple `query` functions.
  * Two constructors 
    *  `  ivf_index(size_t dim, size_t nlist, size_t max_iter,  float tol = 0.000025` initializes an empty `ivf_index`.  The index is filled in the subsequent calls to `train()` and `add()`.  The index can be saved to a group uri with a call to `write()`
    * `ivf_index(tiledb::Context& ctx, const std::string& group_uri)` reads an index that has been stored to a TilDB array, indicated by the `group_uri` argument.
  * `train(const ColMajorMatrix<T>& training_set,  kmeans_init init)` takes a training set of vectors and uses kmeans clustering to find a set of centroids (the cardinality of which is specified on initializing the index).
  * `add(const ColMajorMatrix<T>& training_set)` creates a partitioned inverted file index using previously computed centroids.  Currently, the data comprising the index is held completely in memory.  "Infinite RAM" queries can be applied at this point.
  * Infinite RAM queries: `query_x(const Q& query_vectors, size_t k_nn, size_t nprobe)` queries the index for the `query_vectors`, using parameters `k_nn` and `nprobe`.  
    * The available `infinite_ram` queries are the ones that were previously used as free functions in `ivf/qv.h`: `query_infinite_ram`, `qv_query_heap_infinite_ram`, `nuv_query_heap_infinite_ram`, `nuv_query_heap_infinite_ram_reg_blocked`.  The function `query_infinite_ram` is the one that should be called, the others are there to do performance comparisons / regression testing.
  * FInite RAM queries `query_x(const Q& query_vectors, size_t k_nn, size_t nprobe, size_t upper_bound)` includes the additional parameter `upper_bound` to specify a maximum amount of RAM to use (in terms of vectors) when performing a query.
    * The available `finite_ram` queries are the ones that were previously used as free functions in `ivf/qv.h`: `query_finite_ram`,  `nuv_query_heap_finite_ram`, and `nuv_query_heap_finite_ram_reg_blocked`.  The function `qv_query_heap_finite_ram` is still WIP.  The function `query_finite_ram` is the one that should be called, the others are there to do performance comparisons / regression testing.
  * A member `write_index(const std::string& group_uri, bool overwrite)` that saves the index to `group_uri`.  The group schema currently in place is transitional.  It should be changed to be compatible with our currently used group schema.
  * A number of methods for assisting with testing and debugging.

### Unit tests 

For `ivf_index` and other new classes in this PR.
* Primarily based using `siftsmall` datasets
* `ivf_index` successfully passes
  * unit tests for kmean clustering
  * unit tests to create indexes
  * unit tests to write and read back indexes
  * unit tests to query ivf indexes (all of the queries described above), using the siftsmall dataset.
* because of the interface changes, the unit tests for ivf_qv and ivf_vq have been disabled and will be migrated to the new interface (and perhaps incorporated directly into unit_ivf_index.cc).
* Integration tests with CLI programs give correct results for `siftsmall`, `sift`, `1M`, and `10M` datasets.


NikolaosPapailiou (approved):
Please fix the test

PR #219, opened 2024-01-26
Enable Windows CI/CD jobs

dudoslav:
This PR adds `windows-2022` into the CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,10 +10,11 @@ jobs:
   run-tests:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-2022]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Install OpenBLAS
+        if: ${{ matrix.os == "ubuntu-latest" }}
         run: sudo apt install libopenblas-dev
       - uses: actions/checkout@v3
       - name: Configure CMake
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -10,10 +10,11 @@ jobs:
   run-tests:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-latest, windows-2022]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Install OpenBLAS
+        if: ${{ startsWith(matrix.os, 'ubuntu-') == true }}
```
teo-tsirpanis, line 17:
```suggestion
        os: [ubuntu-latest, windows-latest]
    runs-on: ${{ matrix.os }}
    steps:
      - name: Install OpenBLAS
        if: ${{ runner.os == 'Linux' }}
```

teo-tsirpanis:
This function does memory-mapped I/O using POSIX APIs. It will have to be ported to Windows.

https://github.com/TileDB-Inc/TileDB-Vector-Search/blob/d60424de7746dfe2b5ca5d0406b4b00422bab511/src/include/detail/linalg/tdb_io.h#L377-L436

dudoslav:
> This function does memory-mapped I/O using POSIX APIs. It will have to be ported to Windows.

I already asked @lums658 for this, but I managed to create a small snippet:
https://github.com/TileDB-Inc/TileDB-Vector-Search/commit/df1bfda362b90dc3efaf4f1e6e7180c1ead5d5eb

however, I need to figure out where to get the required files and also implement some tests

ihnorton:
@lums658 @dudoslav AFAICT there's no memory map support in std::filesystem, so we should make separate functions (platform conditional) and fall back to fstream if mmap is not available.

PR #220, opened 2024-01-26
Bring tests up to date

lums658:
As PRs 210 et al continue towards the goal of #173/#185, some of the tests also evolved but we haven't been maintaining that.  This PR simply updates the tests as of PR #216.  

Once PR #216 is merged, this should be a fairly small PR.


src/include/test/unit_ivf_flat_group.cc
```diff
@@ -160,6 +158,7 @@ TEST_CASE("ivf_flat_group: write constructor - create", "[ivf_flat_group]") {
 
 TEST_CASE(
     "ivf_flat_group: write constructor - create and open", "[ivf_flat_group]") {
+  bool debug = false;
```
jparismorgan, line 161: (Minor) It looks like this is unused, can we removed?

src/include/test/unit_ivf_flat_group.cc
```diff
@@ -179,6 +178,7 @@ TEST_CASE(
 
 TEST_CASE(
     "ivf_flat_group: write constructor - create and read", "[ivf_flat_group]") {
+  bool debug = false;
```
jparismorgan, line 181: (Minor) It looks like this is unused, can we removed?

src/include/test/unit_ivf_flat_group.cc
```diff
@@ -199,6 +199,7 @@ TEST_CASE(
 TEST_CASE(
     "ivf_flat_group: write constructor - create, write, and read",
     "[ivf_flat_group]") {
+  bool debug = false;
```
jparismorgan, line 202: (Minor) It looks like this is unused, can we removed?

src/include/test/unit_queries.cc
```diff
@@ -113,6 +113,7 @@ TEST_CASE("test queries", "[queries]") {
       }
     }
 
+#ifdef TILEDB_VS_ENABLE_BLAS
```
jparismorgan, line 116: Why do we need this? Wasn't the test passing before without BLAS?

src/include/test/unit_ivf_flat_group.cc
lums658: Hmm.  We may be out of sync.  All of the declared `debug` are used in the current version of the branch.

src/include/test/unit_queries.cc
ihnorton: x-ref https://app.shortcut.com/tiledb-inc/story/40724

ihnorton (approved):
LGTM

src/include/test/unit_queries.cc
jparismorgan: Don't quite follow, doesn't that issue suggest we should not be adding more things referring to BLAS? Or are you saying that `gemm_query()` only works with BLAS?

src/include/test/unit_ivf_flat_group.cc
jparismorgan: Can you share what code you are looking at? In the code in this PR I don't see any references to `debug` in this `TEST_CASE`:

Screenshot 2024-02-09 at 11 22 06 AM

PR #223, opened 2024-01-30
Add a test that verifies there are no output logs

jparismorgan:
### What
We add a test that verifies there are no output logs. This should prevent stray debug logs from being introduced.

### Testing
* Unit tests pass.
* If I add a log then the test fails.

### DIFF
--- a/apis/python/test/conftest.py
+++ b/apis/python/test/conftest.py
@@ -1,0 +1,18 @@
+import pytest
+import ctypes
+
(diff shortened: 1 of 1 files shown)


ihnorton (approved):
Thanks! We [had problems](https://github.com/TileDB-Inc/TileDB-Py/pull/601#issuecomment-865264461) with capfd in the past, so just as a heads-up that we might need to disable it on Windows (or debug) if we see weird behavior on Windows.

PR #224, opened 2024-01-30
Implement read_bin_local using VFS

dudoslav:
This PR aims to implement read_bin_local using fstream instead of mmap, since mmap is not supported outside of unix platforms e.g. windows builds fail.

### DIFF
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -37,7 +37,6 @@
 #include <vector>
 
 #include <tiledb/tiledb>
-// #include <tiledb/vfs.h>
 #include "detail/linalg/matrix.h"
 #include "detail/linalg/tdb_helpers.h"
 #include "utils/logging.h"
@@ -378,7 +377,7 @@ auto read_bin_local(
     size_t subset = 0) {
   auto vfs = tiledb::VFS{ctx};
 
-  std::uint32_t dimension = 0u;
+  uint32_t dimension = 0u;
 
   const auto file_size = vfs.file_size(bin_file);
   auto filebuf = tiledb::VFS::filebuf{vfs};
@@ -406,8 +405,8 @@ auto read_bin_local(
   auto result = ColMajorMatrix<T>{dimension, num_vectors};
   auto* result_ptr = result.data();
 
-  for (std::size_t i = 0; i < num_vectors; ++i) {
-    std::uint32_t d = 0u;
+  for (size_t i = 0; i < num_vectors; ++i) {
+    uint32_t d = 0u;
 
     if (!file.read(reinterpret_cast<char*>(&d), sizeof(d))) {
       throw std::runtime_error(
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -481,7 +481,7 @@ auto read_bin_local(const std::string& bin_file, size_t subset = 0) {
 
     if (!file.read(reinterpret_cast<char*>(&d), sizeof(d))) {
       throw std::runtime_error(
-        "failed to read dimension for vector at pos: " + std::to_string(i));
+          "failed to read dimension for vector at pos: " + std::to_string(i));
     }
 
     if (d != dimension) {
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -32,10 +32,26 @@
 #ifndef TILEDB_TDB_IO_H
 #define TILEDB_TDB_IO_H
 
+#if defined __has_include
+#if __has_include (<sys/mman.h>)
+
+#define USE_MMAP
 #include <fcntl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
 #include <unistd.h>
+
+#else
+
+#include <fstream>
+
+#endif
+#else
+
+#include <fstream>
+
+#endif
+
 #include <filesystem>
 #include <numeric>
 #include <vector>
@@ -376,6 +392,7 @@ auto sizes_to_indices(const std::vector<T>& sizes) {
  */
 template <class T>
 auto read_bin_local(const std::string& bin_file, size_t subset = 0) {
+#ifdef USE_MMAP
   if (!std::filesystem::exists(bin_file)) {
     throw std::runtime_error("file " + bin_file + " does not exist");
   }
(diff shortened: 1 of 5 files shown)


src/include/detail/linalg/tdb_io.h
```diff
@@ -376,63 +415,40 @@ auto sizes_to_indices(const std::vector<T>& sizes) {
  */
 template <class T>
 auto read_bin_local(const std::string& bin_file, size_t subset = 0) {
-  if (!std::filesystem::exists(bin_file)) {
-    throw std::runtime_error("file " + bin_file + " does not exist");
-  }
```
jparismorgan, line 381: (Minor) What's the reasoning to remove?

src/include/detail/linalg/tdb_io.h
```diff
@@ -376,63 +415,40 @@ auto sizes_to_indices(const std::vector<T>& sizes) {
...
-  if (fd == -1) {
-    throw std::runtime_error("could not open " + bin_file);
-  }
+  std::uint32_t dimension = 0u;
 
-  uint32_t dimension{0};
-  auto num_read = read(fd, &dimension, 4);
-  lseek(fd, 0, SEEK_SET);
+  const auto file_size = std::filesystem::file_size(bin_file);
+  auto file = std::ifstream{bin_file, std::ios::binary};
```
jparismorgan, line 421: (Unsure) Do we need to add a `if (!file.is_open()) {` check? I see they do that here: https://en.cppreference.com/w/cpp/io/basic_ifstream

src/include/detail/linalg/tdb_io.h
```diff
@@ -376,63 +415,40 @@ auto sizes_to_indices(const std::vector<T>& sizes) {
...
+    if (d != dimension) {
       throw std::runtime_error(
-          "dimension mismatch: " + std::to_string(dim) +
+          "dimension mismatch: " + std::to_string(d) +
           " != " + std::to_string(dimension));
     }
-    std::copy(sift_ptr, sift_ptr + dimension, data_ptr);
-    data_ptr += dimension;
-    sift_ptr += dimension;
+    if (not file.read(reinterpret_cast<char*>(result_ptr), d * sizeof(T))) {
```
jparismorgan, line 445: (Minor)(Optional) Does `clang-format` not fail on this? I would have expected:
```suggestion
    if (!file.read(reinterpret_cast<char*>(result_ptr), d * sizeof(T))) {
```

jparismorgan (commented):
Generally looks good to me, just small issues.

src/include/detail/linalg/tdb_io.h
jparismorgan: Oh wait, it's failing. Will be curious to see once you run it.

src/include/detail/linalg/tdb_io.h
dudoslav: The std::fstream{} constructor already checks if that file exists, so I thought that this code is redundant, meaning it would be checked twice. (No one guarantees that between filestystem::exists and opening the file, the file would not be deleted). File paths are not atomic.

src/include/detail/linalg/tdb_io.h
dudoslav: I think that we should add check for each read. The read would inevitably fail if the file failed to open

src/include/detail/linalg/tdb_io.h
dudoslav: :( I am used to writing it like this, but my bad I will change it to `!` .

src/include/detail/linalg/tdb_io.h
jparismorgan: Sorry, don't follow. You're saying the caller should check something after calling `read_bin_local()`?

dudoslav: Ah sorry my bad. I assumed that the `std::iostream` constructor would throw an exception (and it actually does) so it would never get to checking `file.open` if it failed to open that file. But it would not hurt to add this check so I'll do it.

dudoslav:
I cleaned this PR so that it could be merged with master. Also as for @ihnorton 's request I kept the original implementation for platforms that support `mmap`. Let me know if I did it correctly, and if it is really necessary.

src/include/detail/linalg/tdb_io.h
```diff
@@ -32,10 +32,26 @@
...
+#else
+
+#include <fstream>
+
+#endif
+#else
+
+#include <fstream>
+
+#endif
```
teo-tsirpanis, line 53:
```suggestion
#if defined __has_include
#if __has_include(<sys/mman.h>)
#define USE_MMAP
#endif
#endif

#ifdef USE_MMAP
#include <fcntl.h>
#include <sys/mman.h>
#include <sys/stat.h>
#include <unistd.h>
#else
#include <fstream>
#endif
```

Shelnutt2:
Why not use VFS so we can universally support all storage backends?

dudoslav:
> Why not use VFS so we can universally support all storage backends?

Could you elaborate? Maybe provide some links?

src/include/detail/linalg/tdb_io.h
```diff
@@ -32,15 +32,12 @@
...
-#include <fcntl.h>
-#include <sys/mman.h>
-#include <sys/stat.h>
-#include <unistd.h>
 #include <filesystem>
 #include <numeric>
 #include <vector>
 
 #include <tiledb/tiledb>
+// #include <tiledb/vfs.h>
```
jparismorgan, line 40: (Minor) Remove?

src/include/detail/linalg/tdb_io.h
```diff
@@ -375,64 +372,60 @@ auto sizes_to_indices(const std::vector<T>& sizes) {
...
-    throw std::runtime_error("file " + bin_file + " does not exist");
-  }
-  auto file_size = std::filesystem::file_size(bin_file);
+auto read_bin_local(
+    const tiledb::Context& ctx,
+    const std::string& bin_file,
+    size_t subset = 0) {
+  auto vfs = tiledb::VFS{ctx};
+
+  std::uint32_t dimension = 0u;
```
jparismorgan, line 381: (Minor) Do we need `std::` here? We don't do this in the rest of the codebase. Same below.

jparismorgan (approved):
Looks good to me with two small nits.

ihnorton:
cc @lums658

lums658 (approved):
Looks good to me.

PR #225, opened 2024-01-31
Reintegrate flatpq index

lums658:
This PR moves over the initial implementation of the `flatpq_index` class as originally implemented in lums/[sc-32521](https://app.shortcut.com/tiledb-inc/story/32521)/pq and in PR #153.

### Product Quantization

Product quantization (pq) is a method for compressing vectors in a vector database.  The basic idea is to divide the vectors into some number of subvectors and then to use the nearest centroid to each subvector to represent the original vector.  Since there are a fixed number of actual centroids, we only need to store the index of the centroid.  
For example, if we have 256 centroids, we can represent each centroid with an 8-bit index.  If we subdivide a vector into, say, eight subvectors, we can represent each subvector with 8bits, for a total of 64bits for the entire vector.  If the original vector comprised 128 floats, this is a compression of (128 * 32 ) / 64 -- a factor of 64. 

In addition to space savings, distance computation can also be greatly accelerated.  If we are using the "L2" distance -- which is really the sum of squares -- the distance between two vectors is the sum of the distance between each of its subvectors.  Since each subvector is represented by the index of the nearest centroid, we can store those distances in a lookup table and access them with a simple lookup.

The tutorial at https://www.pinecone.io/learn/series/faiss/product-quantization/ explains the general idea of product quantization in more detail.

### `flatpq_index`

To realize PQ in TileDB-Vector-Search, we implement the following:
* The `flatpq_index` constructor defines the number of subvectors and bits per subvector (equivalent to the number of centroids).
* `train()` runs a kmeans algorithm to compute the centroids for each subspace.  Each subspace has its own set of centroids.  For each subspace, a 2D table is created holding the distance between each pair of centroids in the subspace.
* `add()` takes a set of input vectors and runs `qv_partition` on each subspace to obtain the codes.
* `symmetri_query()` first encodes the query vectors and then calls `flat::qv_query_heap` on the encoded queries, using the symmetric pq distance function defined in the class.  They symmetric distance function uses the lookup tables for computing distances.
* `asymmetric_query()` calls `flat::qv_query_heap` directly on the query vectors, using the asymmetric pq distance function defined in the class.  The asymmetric distance function decodes the encoded vector and computes the distance in the normal fashion.
* `write_index` saves the index information.  This needs to be brought up to speed by defining index group and index metadata classes.  
* The "loading" constructor reads an index -- it also needs to be brought up to speed.

### DIFF
--- a/src/include/index/flatpq_index.h
+++ b/src/include/index/flatpq_index.h
@@ -1,0 +1,927 @@
+/**
+ * @file   flatpq_index.h
+ *
+ * @section LICENSE
+ *
+ * The MIT License
+ *
+ * @copyright Copyright (c) 2023 TileDB, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @section DESCRIPTION
+ *
+ * Header-only library of class that implements a flat index that used product
+ * quantization.
+ *
+ */
+
+#ifndef TILEDB_FLATPQ_INDEX_H
+#define TILEDB_FLATPQ_INDEX_H
+
+#include <cstddef>
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <unordered_set>
+
+#include "detail/flat/qv.h"
+#include "detail/linalg/matrix.h"
+#include "detail/linalg/tdb_io.h"
+
+#include <tiledb/group_experimental.h>
+#include <tiledb/tiledb>
+
+template <class T, class U>
+void sub_kmeans_random_init(
+    const ColMajorMatrix<T>& training_set,
+    ColMajorMatrix<U>& centroids,
+    size_t sub_begin,
+    size_t sub_end,
+    size_t seed = 0) {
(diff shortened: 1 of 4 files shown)


src/include/index/flatpq_index.h
```diff
@@ -0,0 +1,927 @@
...
+ * @return
+ *
+ * @todo update with concepts
+ * @todo fix up zero sized partitions
+ * @todo this would be more cache friendly to do sub_begin and sub_end
+ * in an inner loop
+ * @todo We can probably just reuse plain kmeans at some point.
+ */
+template <class T, class U>
+auto sub_kmeans(
```
NikolaosPapailiou, line 111: Can we use the existing kmeans library for this task?

src/include/index/flatpq_index.h
```diff
@@ -0,0 +1,927 @@
...
+    for (size_t j = 0; j < dimension(new_centroids); ++j) {
+      new_centroids(j, i) = centroids(j, i);
+    }
+  }
+
+  size_t iter = 0;
+  double max_diff = 0.0;
+  double total_weight = 0.0;
+  for (iter = 0; iter < max_iter; ++iter) {
+#ifdef REASSIGN
```
NikolaosPapailiou, line 142: Should this be a function argument rather than #ifdef?

src/include/index/flatpq_index.h
```diff
@@ -0,0 +1,927 @@
...
+          "dimension, num_subspaces, and bits_per_subspace must be greater "
+          "than zero");
+    }
+    if ((dimension_ % num_subspaces_) != 0) {
+      throw std::invalid_argument(
+          "Number of subspaces must evenly divide dimension of vector");
+    }
+    sub_dimension_ = dimension_ / num_subspaces_;
+
+#if 0
```
NikolaosPapailiou, line 350: Please cleanup #ifs

src/include/index/flatpq_index.h
```diff
@@ -0,0 +1,927 @@
...
+      {"bits_per_subspace", &bits_per_subspace_, TILEDB_UINT64},
+      {"num_clusters", &num_clusters_, TILEDB_UINT64},
+      {"tol", &tol_, TILEDB_FLOAT32},
+      {"max_iter", &max_iter_, TILEDB_UINT64},
+      {"num_threads", &num_threads_, TILEDB_UINT64},
+  };
+
+  // array data
+  ColMajorMatrix<centroid_feature_type> centroids_;
+  std::vector<ColMajorMatrix<score_type>> distance_tables_;
```
NikolaosPapailiou, line 316: What is the expected RAM size for this?

NikolaosPapailiou (approved):

src/include/index/flatpq_index.h
lums658: Not very much: (score size) * (# centroids)^2 * (# subspaces).  If we take the usual float scores, 8-bit codes, and 8 subspaces, that is 4 * 256 * 256 * 8 = 2MB -- for all of the tables.  Each table is 256KB.

src/include/index/flatpq_index.h
lums658: We should definitely integrate them as much as we can.  The main difference between the two is whether or not to operate on a subset of the vectors or on the whole vector (there are some small differences as well).  Since the whole vector is just a particular subset, the same code should be usable.  I have created a story for this.

src/include/index/flatpq_index.h
lums658: Yes, although we can't directly substitute an `if` statement for the `ifdef` as there are some structural / scoping issues that have to be handled.  Deferring to future PR (sc-40508).

jparismorgan (approved):

PR #226, opened 2024-02-02
Add directory document readers for vector search Object API

NikolaosPapailiou:
This is adding directory readers that can be used to read documents from a directory structure. 
We introduce 2 readers:

1. Text reader: Extracts text from different types of documents based on their mime type. This can be used for several file types (.pdf, .docx, .html, .jpg, .png, etc.). This is implemented by reusing the LangChain document read utilities.
2. Image document reader: Read images from image files in a directory. 

The readers also work for vfs directories.

We also introduce multiple notebook examples in `apis/python/examples/object_api/`


jparismorgan (approved):
Looks good to me but you may want to wait for another review as well.

PR #229, opened 2024-02-06
Add type-erased vamana index

jparismorgan:
### What
Here we add a type-erased `IndexVamana` which can be used from Python. For `src/include/api/vamana_index.h` we follow the pattern of `src/include/api/ivf_flat_index.h`.

### Testing
* Adds a unit test to construct, train, and query a `IndexVamana`.

### DIFF
--- a/src/include/api/vamana_index.h
+++ b/src/include/api/vamana_index.h
@@ -87,7 +87,6 @@ class IndexVamana {
    */
   explicit IndexVamana(
       const std::optional<IndexOptions>& config = std::nullopt) {
-    std::cout << "[IndexVamana@1] Opening index" << std::endl;
     feature_datatype_ = TILEDB_ANY;
     id_datatype_ = TILEDB_UINT32;
     px_datatype_ = TILEDB_UINT32;
@@ -96,7 +95,6 @@ class IndexVamana {
       for (auto&& c : *config) {
         auto key = c.first;
         auto value = c.second;
-        std::cout << "[IndexVamana@1]" << key << " " << value << std::endl;
         if (key == "dimension") {
           dimension_ = std::stol(value);
         } else if (key == "L_build") {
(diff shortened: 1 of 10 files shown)


src/include/detail/linalg/matrix.h
```diff
@@ -70,6 +70,14 @@ class MatrixView : public stdx::mdspan<T, matrix_extents<I>, LayoutPolicy> {
       : Base{p, r, c} {
   }
 
+  auto data() {
+    return this->data_handle();
+  }
+
+  auto data() const {
+    return this->data_handle();
+  }
```
jparismorgan, line 79: This was changed because in `src/include/api/vamana_index.h` we do:
```
void train(const FeatureVectorArray& training_set) override {
      using feature_type = typename T::value_type;
      auto fspan = MatrixView<feature_type, stdx::layout_left>{
          (feature_type*)training_set.data(),
          extents(training_set)[0],
          extents(training_set)[1]};
      impl_index_.train(fspan);
      ...
```
which calls `train()` in `vamana.h`, which then calls `.data()`:
```
template <feature_vector_array Array>
void train(const Array& training_set) {
    feature_vectors_ = std::move(ColMajorMatrix<feature_type>(
        ::dimension(training_set), ::num_vectors(training_set)));
    std::copy(
        training_set.data(),
        training_set.data() +
            ::dimension(training_set) * ::num_vectors(training_set),
        feature_vectors_.data());
    ...
```
leading to this build failure:
```
/Users/parismorgan/repo/TileDB-Vector-Search-4/src/include/detail/graph/vamana.h:531:22: error: no member named 'data' in 'MatrixView<float, Kokkos::layout_left>'
    training_set.data(),
```

NikolaosPapailiou (approved):

src/include/detail/linalg/matrix.h
lums658: I think it is probably safe to add a `data()` member to `MatrixView`.  We would need to add a data member to record the pointer passed in by the constructor since `mdspan` does not have a `data()` or `get()` member function.  We definitely need `MatrixView` to provide a non-owning lightweight wrapper around contiguous storage.

src/include/api/vamana_index.h
```diff
@@ -0,0 +1,498 @@
...
+    dimension_ = index_->dimension();
+  }
+
+  /**
+   * @brief Train the index based on the given training set.
+   * @param training_set
+   * @param init
+   */
+  // @todo -- infer feature type from input
+  void train(const FeatureVectorArray& training_set) {
```
lums658, line 225: For a future PR, swap `train` and `add` per our discussion today.  The C++ implementation will need to be swapped as well.  (Or we could go ahead and do it for this PR.)

src/include/detail/graph/vamana.h
```diff
@@ -764,12 +764,13 @@ class vamana_index {
    * those will presumably be in a known array that can be made part of
    * the group?
    */
-  auto write_index(const std::string& group_uri, bool overwrite = false) {
-    // copilot ftw!
+  auto write_index(
+      const tiledb::Context& ctx,
```
lums658, line 768: Ah.  Yes.

lums658 (approved):

PR #232, opened 2024-02-12
Add an API for `externals_ids` to `IndexVamana`

jparismorgan:
### What
Follows the pattern of `IndexIVFFlat` and `IndexFlatL2` and adds a `add_with_ids` method to `IndexVamana`, as well as a `train_with_ids` method. The implementation is not done, but opening up this PR to see if we like this API. Next step will be the implementation.

Note that I actually wanted the API to be this:
```
.def("train",
    [](IndexVamana& index, const FeatureVectorArray& vectors, const std::optional<FeatureVectorArray>& external_ids) {
            index.train(vectors, external_ids);
          },
          py::arg("vectors"),
          py::arg_v("external_ids", std::nullopt, "None"))
```
But when I do that:
```
TypeError: train(): incompatible function arguments. The following argument types are supported:
       1. (self: tiledb.vector_search._tiledbvspy.IndexVamana, vectors: tiledb.vector_search._tiledbvspy.FeatureVectorArray, external_ids: std::__1::optional<FeatureVectorArray> = None) -> None
Invoked with: <tiledb.vector_search._tiledbvspy.IndexVamana object at 0x10efbc330>, <tiledb.vector_search._tiledbvspy.FeatureVectorArray object at 0x10efbc370>, <tiledb.vector_search._tiledbvspy.FeatureVectorArray object at 0x10efbc370>
```

So I figured I might as well keep moving forward, and if I figure out how to do this I can update the API.

Note that we also update `IndexVamana@query` to take in a `std::optional<size_t> opt_l`.

### Testing
* Adds a unit test.

### DIFF
--- a/apis/python/src/tiledb/vector_search/type_erased_module.cc
+++ b/apis/python/src/tiledb/vector_search/type_erased_module.cc
@@ -254,10 +254,18 @@ void init_type_erased_module(py::module_& m) {
             auto args = kwargs_to_map(kwargs);
             new (&instance) IndexVamana(args);
           })
-      // TODO(paris): Update train()/add() to take `const std::optional<FeatureVectorArray>& external_ids`. Failing with this currently:
-      // TypeError: train(): incompatible function arguments. The following argument types are supported:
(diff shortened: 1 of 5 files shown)


jparismorgan:
Actually will pause on this for now and re-open later.

PR #234, opened 2024-02-12
Draft: Fix Path to String conversions

dudoslav:
This PR demonstrates how we could fix all path to string conversions since windows uses a different underlying string container for `std::filesystem::path` class.

This PR is just a draft, if this is something we want to include in the next release I will clean it up and simplify all changes.

### DIFF
--- a/src/include/api/feature_vector_array.h
+++ b/src/include/api/feature_vector_array.h
@@ -68,6 +68,12 @@ class FeatureVectorArray {
     feature_size_ = datatype_to_size(feature_type_);
   }
 
+    FeatureVectorArray(
+      const tiledb::Context& ctx,
+      const std::filesystem::path& uri,
+      size_t num_vectors = 0) : FeatureVectorArray(ctx, uri.string(), num_vectors) {
+    }
+
   FeatureVectorArray(
       const tiledb::Context& ctx,
       const std::string& uri,
--- a/src/include/index/index_group.h
+++ b/src/include/index/index_group.h
@@ -130,8 +130,8 @@ class base_index_group {
     }
     for (auto&& [array_key, array_name] : storage_formats[version_]) {
       valid_key_names_.insert(array_key);
-      valid_array_names_.insert(array_name);
-      array_name_map_[array_key] = array_name;
+      valid_array_names_.insert(array_name.string());
+      array_name_map_[array_key] = array_name.string();
     }
     static_cast<group_type*>(this)->append_valid_array_names_impl();
   }
@@ -166,11 +166,12 @@ class base_index_group {
    */
   void init_for_open(const tiledb::Config& cfg) {
     tiledb::VFS vfs(cached_ctx_);
-    if (!vfs.is_dir(group_uri_)) {
+    if (!vfs.is_dir(group_uri_.string())) {
       throw std::runtime_error(
-          "Group uri " + std::string(group_uri_) + " does not exist.");
+          "Group uri " + group_uri_.string() + " does not exist.");
     }
-    auto read_group = tiledb::Group(cached_ctx_, group_uri_, TILEDB_READ, cfg);
+    auto read_group =
+        tiledb::Group(cached_ctx_, group_uri_.string(), TILEDB_READ, cfg);
 
     // Load the metadata and check the version.  We need to do this before
     // we can check the array names.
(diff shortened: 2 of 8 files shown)


jparismorgan:
I'm happy to move forward with this, looks like tests pass and the changes seem nice.

src/include/test/unit_api_feature_vector_array.cc
```diff
@@ -46,22 +46,22 @@ TEST_CASE("api_feature_vector_array: test test", "[api_feature_vector_array]") {
 TEST_CASE("api: feature vector array open", "[api]") {
   tiledb::Context ctx;
 
-  auto a = FeatureVectorArray(ctx, sift_inputs_uri);
+  auto a = FeatureVectorArray(ctx, sift_inputs_uri.string());
```
lums658, line 49: Since `FeatureVectorArray` has an overload for `std::filesystem::path`, we should use that instead of converting to a `string` here.  Similarly for other classes such as `FlatL2Index` (see comment above).

src/include/api/feature_vector_array.h
```diff
@@ -68,6 +68,12 @@ class FeatureVectorArray {
     feature_size_ = datatype_to_size(feature_type_);
   }
 
+    FeatureVectorArray(
```
lums658, line 71: We should add overloads like this to the other type-erased api classes, e.g., `FlatL2Index`, `IVFFlatIndex`.  This would remove the need to explicitly convert to string at the call site.  See comments below.

The basic idea is that everything should work *as if* the automatic conversion from `std::filesystem::path` to `std::string` actually worked.

Some other classes might need overloads then as well -- `tdbMatrix` for instance.

If adding those overloads bloats the code too much, then we can revisit this and maybe leave some that don't have the overloads.

src/include/index/index_group.h
```diff
@@ -166,11 +166,12 @@ class base_index_group {
    */
   void init_for_open(const tiledb::Config& cfg) {
     tiledb::VFS vfs(cached_ctx_);
-    if (!vfs.is_dir(group_uri_)) {
+    if (!vfs.is_dir(group_uri_.string())) {
```
lums658, line 169: Just an aside -- TileDB core should handle `std::filesystem::path` transparently.

src/include/test/unit_api_feature_vector_array.cc
```diff
@@ -199,25 +199,26 @@ TEST_CASE("api: query checks", "[api][index]") {
...
+    auto z = FeatureVectorArray(ctx, sift_inputs_uri.string());
     auto nn = dimension(z);
     auto nnn = num_vectors(z);
     CHECK(dimension(z) == 128);
     CHECK(num_vectors(z) == num_sift_vectors);
   }
 
   SECTION("tdbMatrix") {
-    auto ck = tdbColMajorMatrix<float>(ctx, sift_inputs_uri);
+    auto ck = tdbColMajorMatrix<float>(ctx, sift_inputs_uri.string());
```
lums658, line 210: We should probably also have an overload with `std::filesystem::path` for `tdbColMajorMatrix` etc.

src/include/test/unit_api_flat_l2_index.cc
```diff
@@ -43,7 +43,7 @@ TEST_CASE("api_flat_l2_index: test test", "[api_flat_l2_index]") {
 
 TEST_CASE("api: flat_l2_index", "[api][flat_l2_index]") {
   tiledb::Context ctx;
-  auto a = IndexFlatL2(ctx, fmnist_inputs_uri);
+  auto a = IndexFlatL2(ctx, fmnist_inputs_uri.string());
```
lums658, line 46: Again, we should have an overload, rather than requiring explicit conversion to string.

lums658 (changes requested):
The main comment is that we should have more overloads of classes so that they can take a `std::filesystem::path` in addition to taking a `std::string`.  User code then doesn't have to convert to a string at the call site.  (Basically, we want our api to behave *as if* automatic conversion was working.)

For some of our internal utility functions we probably don't need this, but for user-facing api code we do.

src/include/api/feature_vector_array.h
dudoslav: I completely agree, I will have to clean this PR

dudoslav:
Closing this PR in favor of another PR that removes use of filesystem::path

PR #239, opened 2024-02-15
Enable PYPI upload workflow without Windows packages

dudoslav:
This PR implements PyPi upload for linux and macosx platforms since windows pipeline still fails. Source distribution is available as well.

The previous PR that fails and is postponed is here: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/207

Edit: Recreate PR from this repo instead of fork so that all tests can pass

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -5,12 +5,12 @@ on:
   push:
     branches:
       - release-*
-      - "*wheel*" # must quote since "*" is a YAML reserved character; we want a string
+      - '*wheel*'  # must quote since "*" is a YAML reserved character; we want a string
     tags:
-      - "*"
+      - '*'
   pull_request:
     branches:
-      - "*wheel*" # must quote since "*" is a YAML reserved character; we want a string
+      - '*wheel*'  # must quote since "*" is a YAML reserved character; we want a string
 
 jobs:
   generate_backwards_compatibility_data:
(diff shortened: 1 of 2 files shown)


ihnorton:
@jparismorgan please give this a once-over and merged if ok w/ you.

.github/workflows/build-wheels.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           echo $release_tag
 
           # Install dependencies.
-          pip install .
+          cd apis/python && pip install . && cd ../..
```
jparismorgan, line 47: Was this a merge conflict? I believe we need to keep this as it was.

jparismorgan (approved):
Approving with the assumption that the `pip install .` change was an accident and is reverted before merge. If that's the case then looks good to me.

.github/workflows/build-wheels.yml
dudoslav: https://github.com/dudoslav/TileDB-Vector-Search/actions/runs/7916707804/job/21611139539#step:6:7

but it seems that it works

dudoslav: I think that that pipeline would not work on release

dudoslav: Returned

jparismorgan (approved):

PR #240, opened 2024-02-15
Add a `tdbMatrixWithIds` class

jparismorgan:
### What
This is a follow-up to #235. To support external ids we want to have `FeatureVectorArray` hold a `MatrixWithIds` or a `TdbMatrixWithIds`. `TdbMatrixWithIds` should take in a URI pointing to IDs and load in the IDs into `MatrixWithIds` as the vectors are loaded in.

### Testing
* Adds a unit test - we follow the same pattern as `unit_tdb_matrix.cc`.


src/include/detail/linalg/matrix_with_ids.h
```diff
@@ -0,0 +1,266 @@
+/**
+ * @file   matrix.h
```
lums658, line 2: All comments in PR #235 apply.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -54,9 +54,13 @@
  * it is sufficient to simply have one Matrix class and have a factory that
  * creates them by reading from TileDB.
  */
-template <class T, class LayoutPolicy = stdx::layout_right, class I = size_t>
-class tdbBlockedMatrix : public Matrix<T, LayoutPolicy, I> {
-  using Base = Matrix<T, LayoutPolicy, I>;
+template <
```
lums658, line 57: Hmm.  For the same reasons that `Matrix` should be standalone and independent of `MatrixWithIds`, `tdbMatrix` and `tdbMatrixWithIds` should be independent of each other.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -253,12 +257,25 @@ class tdbBlockedMatrix : public Matrix<T, LayoutPolicy, I> {
...
     auto data_ = std::unique_ptr<T[]>(new T[dimension * load_blocksize_]);
 #endif
-
-    Base::operator=(Base{std::move(data_), dimension, load_blocksize_});
+    if constexpr (std::is_same<MatrixBase, Matrix<T, LayoutPolicy, I>>::value) {
+      Base::operator=(Base{std::move(data_), dimension, load_blocksize_});
+    } else {
+      std::cout << "hit WithIDS case" << std::endl;
+#ifdef __cpp_lib_smart_ptr_for_overwrite
+      auto ids_data_ =
```
lums658, line 265: No.

src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -0,0 +1,272 @@
...
+          LayoutPolicy,
+          I,
+          MatrixWithIds<T, LayoutPolicy, I, IdsType>> {
+  using Base = tdbBlockedMatrix<
+      T,
+      LayoutPolicy,
+      I,
+      MatrixWithIds<T, LayoutPolicy, I, IdsType>>;
+  using Base::Base;
+
```
lums658, line 68: As with `Matrix` and `MatrixWithIds`, `tdbMatrixWithIds` needs to own the ids.  The inheritance hierarchy should be the same and the ids belong in the "WithIds" class.

src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -0,0 +1,272 @@
...
+ * Initialized in construction by filling from the TileDB vectors array and the
+ * the TileDB IDs array.
+ */
+template <
+    class T,
+    class LayoutPolicy = stdx::layout_right,
+    class I = size_t,
+    class IdsType = size_t>
+class tdbBlockedMatrixWithIds
+    : public tdbBlockedMatrix<
```
lums658, line 57: Why does this not inherit from `MatrixWithIds`?

lums658 (changes requested):
1. Comments from PR #215 all apply
2. `tdbMatrixWithIds` needs to have the ids, not the `tdbMatrix`.
3. If I neglected to put it in PR #215, the "WithIds" versions should have accessors for the underlying `matrix` and for the underlying `ids`.

src/include/detail/linalg/matrix_with_ids.h
jparismorgan: Thanks, though please note that this PR was a draft and so not ready for review yet.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: I think we should discuss this a bit more, as this will lead to a lot of duplication of code, which has a real cost, i.e. more maintenance and harder to change things. By using this approach we only need to write the additional IDs loading and storage.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: This is not a useful comment, please explain more what you mean by this.

src/include/detail/linalg/tdb_matrix_with_ids.h
jparismorgan: It does, through `tdbMatrixWithIds `. But see comment above - let's discuss on that thread and offline.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: Talked offline and have a plan around this now. Will mark this as resolved and let's discuss any other suggestions in a follow-up comment.

src/include/detail/linalg/tdb_matrix_with_ids.h
jparismorgan: Talked offline and have a plan around this now. Will mark this as resolved and let's discuss any other suggestions in a follow-up set of comments.

src/include/test/unit_tdb_matrix_with_ids.cc
```diff
@@ -0,0 +1,180 @@
...
+  CHECK(X.ids()[0] == offset + 0);
+  CHECK(X.ids()[1] == offset + 1);
+  CHECK(X.ids()[10] == offset + 10);
+
+  auto Y = tdbColMajorMatrixWithIds<TestType, TestType>(
+      ctx, tmp_matrix_uri, tmp_ids_uri);
+  Y.load();
+  CHECK(num_vectors(Y) == num_vectors(X));
+  CHECK(dimension(Y) == dimension(X));
+  CHECK(
```
lums658, line 70: We should also check `CHECK(size(Y.ids()) == size(X.ids()));` and also `CHECK(size(Y.ids()) == Y.num_ids());`

src/include/test/unit_tdb_matrix_with_ids.cc
```diff
@@ -0,0 +1,180 @@
...
+  auto X = ColMajorMatrixWithIds<TestType, TestType, size_t>(Mrows, Ncols);
+  fill_and_write_matrix(
+      ctx, X, tmp_matrix_uri, tmp_ids_uri, Mrows, Ncols, offset);
+  CHECK(X.ids()[0] == offset + 0);
+  CHECK(X.ids()[1] == offset + 1);
+  CHECK(X.ids()[10] == offset + 10);
+
+  auto Y = tdbColMajorMatrixWithIds<TestType, TestType>(
+      ctx, tmp_matrix_uri, tmp_ids_uri);
+  Y.load();
```
lums658, line 67: We should probably add a "PreLoad" version of this -- so many bugs caused by forgetting to call `load()`....  The name starts to become a mouthful though....

In debug mode, we should also write the unloaded matrix data with NaN or some kind of garbage that makes it obvious the `Matrix` hasn't been loaded.

src/include/test/unit_tdb_matrix_with_ids.cc
```diff
@@ -0,0 +1,180 @@
...
+  CHECK(dimension(Y) == dimension(X));
+  CHECK(
+      std::equal(X.data(), X.data() + dimension(X) * num_vectors(X), Y.data()));
+  for (size_t i = 0; i < 5; ++i) {
+    for (size_t j = 0; j < 5; ++j) {
+      CHECK(X(i, j) == Y(i, j));
+    }
+  }
+
+  // Check that we can assign to a matrix
```
lums658, line 145: How should we define inter-operation between a `Matrix` and a `MatrixWithIds`?  Should they be able to assign to / copy each other?  We should define this behavior and add appropriate unit tests here and in `unit_matrix_with_ids.cc`

lums658 (approved):
Following some Slack discussions, this all seems good.

There are a few comments here -- the only to-do is to add a few more `CHECK`s to the unit test.  The other comments starting with "we should..." are things for future PRs -- please make stories for them so they aren't forgotten.

src/include/test/unit_tdb_matrix.cc
```diff
@@ -201,3 +202,66 @@ TEMPLATE_TEST_CASE("tdb_matrix: preload", "[tdb_matrix]", float, uint8_t) {
...
+
+  size_t Mrows = 200;
+  size_t Ncols = 500;
+
+  using T = float;
+  using I = size_t;
+  using LayoutPolicy = stdx::layout_left;
+  using IdsType = uint8_t;
+
+  std::string tmp_matrix_uri = "/tmp/tmp_tdb_matrix";
```
NikolaosPapailiou, line 219: Should we use a more random pattern for the test filenames? 
We can do this consistently in a followup PR.

src/include/test/unit_tdb_matrix_with_ids.cc
jparismorgan: Sure, added a story for it: https://app.shortcut.com/tiledb-inc/story/41374

src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -0,0 +1,272 @@
...
+   * @param temporal_policy The TemporalPolicy to use for reading the array
+   * data.
+   */
+  tdbBlockedMatrixWithIds(
+      const tiledb::Context& ctx,
+      const std::string& uri,
+      const std::string& ids_uri,
+      size_t upper_bound,
+      size_t timestamp = 0)
+    requires(std::is_same_v<LayoutPolicy, stdx::layout_left>)
```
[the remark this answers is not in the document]
jparismorgan: I don't exactly know, but [this comment](https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/235#discussion_r1494905625) by @lums658 about `FeatureVectorArray` likely also applies here:

> In general, stdx::layout_right is not well-tested at all. We might want to audit that and make sure it works everywhere as well as stdx::layout_left. The idea at one point was that a feature vector array should be agnostic as to which layout and that we would use a feature vector array everywhere rather than a ColMajorMatrix. I have started putting concepts in many of the function templates as a start to that.
>
> The layout doesn't really matter, because as long as the underlying one-dimensional data has each of the vectors as contiguous data, both row major and column major will have, respectively, rows and columns as contiguous data. We don't ever iterate over a matrix element by element, but rather operator[] pulls out a row or column, i.e., a contiguous vector.
>
> For the same reason, I think the before and after calls in Python to transpose are superfluous.
>
> In case you are wondering why I picked a column major matrix to represent a feature vector array -- in linear algebra a vector is almost universally represented as a column.

src/include/test/unit_tdb_matrix_with_ids.cc
jparismorgan: So today it looks like we can move from a `MatrixWithIds` to a `Matrix` okay, but not the other way around. What is your preference here on end behavior? It seems like we won't be doing this move often so I would be fine preventing it (if it's possible which I'm guessing it is)?

src/include/test/unit_tdb_matrix.cc
jparismorgan: Sure, added a story for it: https://app.shortcut.com/tiledb-inc/story/41375

src/include/test/unit_tdb_matrix_with_ids.cc
jparismorgan: Thanks, added more checks.

PR #245, opened 2024-02-19
Fixes the `Generate Backwards Compatibility Data` job

jparismorgan:
### What
In this job: https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/7935942796/job/21670159062#step:6:330 we failed with:
```
must provide `--title` and `--body` (or `--fill` or `fill-first`) when not running interactively
```
So here we do that.

### Testing
None, but the error is simple enough.


dudoslav (approved):
Looks good to me, approved! :1st_place_medal:

NikolaosPapailiou (approved):

PR #246, opened 2024-02-20
Replace switch-case statements in type-erased `FeatureVectorArray` API with a lookup table

jparismorgan:
### What
We can improve the code ergonomics by switching to use a lookup table instead of a switch case. In a follow-up we'll also need to add the ID type, but doing it in two parts to make individual reviews easier.

### Testing
Unit tests pass.

### Note
I did try to use `constexpr` as shown in https://app.shortcut.com/tiledb-inc/story/41286/replace-switch-case-statements-in-type-erased-api-with-lookup-tables:
```
static constexpr col_major_matrix_table_type col_major_matrix_dispatch_table;

constexpr FeatureVectorArray::col_major_matrix_table_type
    FeatureVectorArray::col_major_matrix_dispatch_table = {
        {TILEDB_FLOAT32, [](size_t rows, size_t cols) { return std::make_unique<FeatureVectorArray::vector_array_impl<ColMajorMatrix<float>>>(rows, cols);}},
...
``` 
but get:
```
In file included from /Users/parismorgan/repo/TileDB-Vector-Search-2/src/include/test/unit_api_feature_vector_array.cc:32:
/Users/parismorgan/repo/TileDB-Vector-Search-2/src/include/api/feature_vector_array.h:197:48: error: constexpr variable cannot have non-literal type 'const col_major_matrix_table_type' (aka 'const map<tiledb_datatype_t, function<unique_ptr<vector_array_base> (unsigned long, unsigned long)>>')
  static constexpr col_major_matrix_table_type col_major_matrix_dispatch_table;
                                               ^
```
I think this makes sense and is expected to not work, but let me know if I'm missing something here.

### DIFF
--- a/src/include/api/feature_vector_array.h
+++ b/src/include/api/feature_vector_array.h
@@ -84,88 +84,24 @@ class FeatureVectorArray {
      * happen with either orientation, and so will work at the other end with
      * either orientation since we are just passing a pointer to the data.
      */
-    if (tdb_col_major_matrix_dispatch_table.find(feature_type_) == tdb_col_major_matrix_dispatch_table.end()) {
+    if (tdb_col_major_matrix_dispatch_table.find(feature_type_) ==
+        tdb_col_major_matrix_dispatch_table.end()) {
       throw std::runtime_error("Unsupported attribute type");
     }
-    vector_array = tdb_col_major_matrix_dispatch_table.at(feature_type_)(ctx, uri, num_vectors);
-//    switch (feature_type_) {
-//      case TILEDB_FLOAT32:
-//        vector_array =
-//            std::make_unique<vector_array_impl<tdbColMajorMatrix<float>>>(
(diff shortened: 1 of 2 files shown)


src/include/api/feature_vector_array.h
```diff
@@ -246,13 +190,114 @@ class FeatureVectorArray {
   };
 
  private:
+  using col_major_matrix_constructor_function =
+      std::function<std::unique_ptr<vector_array_base>(size_t, size_t)>;
+  using col_major_matrix_table_type =
+      std::map<tiledb_datatype_t, col_major_matrix_constructor_function>;
+  static const col_major_matrix_table_type col_major_matrix_dispatch_table;
```
lums658, line 197: There is no `constexpr std::map` so that should not have been expected to work.

NikolaosPapailiou (approved):

PR #247, opened 2024-02-20
Replace switch-case statements in type-erased `api/vamana_index.h` API with a lookup table

jparismorgan:
### What
A few changes:
1. Replace switch-case statements in type-erased `api/vamana_index.h` API with a lookup table 
2. Add a unit test for `src/include/api/vamana_index.h`
3. While adding this test I found a error because we expect `px_datatype` to be in the metadata. So we add it:
```
IndexVamana(
      const tiledb::Context& ctx,
      const URI& group_uri,
      const std::optional<IndexOptions>& config = std::nullopt) {
    using metadata_element = std::tuple<std::string, void*, tiledb_datatype_t>;
    std::vector<metadata_element> metadata{
        {"feature_datatype", &feature_datatype_, TILEDB_UINT32},
        {"id_datatype", &id_datatype_, TILEDB_UINT32},
        {"px_datatype", &px_datatype_, TILEDB_UINT32}};

    tiledb::Config cfg;
    tiledb::Group read_group(ctx, group_uri, TILEDB_READ, cfg);

    for (auto& [name, value, datatype] : metadata) {
      if (!read_group.has_metadata(name, &datatype)) {
        throw std::runtime_error("Missing metadata: " + name);
      }
```

### Testing
Adds a unit test.

### DIFF
--- a/src/include/api/vamana_index.h
+++ b/src/include/api/vamana_index.h
@@ -62,7 +62,7 @@
  * datatypes:
  *   - feature_type: uint8 or float
  *   - id_type: uint32 or uint64
- *   - px_type: uint32 or uint64
+ *   - adjacency_row_index_type: uint32 or uint64
  */
 class IndexVamana {
  public:
@@ -90,7 +90,7 @@ class IndexVamana {
       const std::optional<IndexOptions>& config = std::nullopt) {
     feature_datatype_ = TILEDB_ANY;
     id_datatype_ = TILEDB_UINT32;
-    px_datatype_ = TILEDB_UINT32;
+    adjacency_row_index_datatype_ = TILEDB_UINT32;
 
     if (config) {
       for (auto&& c : *config) {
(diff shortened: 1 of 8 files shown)


NikolaosPapailiou (approved):

src/include/api/vamana_index.h
```diff
@@ -58,6 +58,11 @@
  *   - An add method
  *   - A query method
  *
+ * We support all combinations of the following types for feature, id, and px
+ * datatypes:
+ *   - feature_type: uint8 or float
+ *   - id_type: uint32 or uint64
+ *   - px_type: uint32 or uint64
```
lums658, line 65: vamana doesn't use a "px" datatype.  I used the name "px" in the context of IVF to be distinct from "index" because "index" is quite overloaded -- "px" means "partitioning index' -- i.e., the array that demarcates the begin and end of each partition in IVF. There is an analogous thing in `vamana_index` -- the type used as an indexing array that is part of vamana's graph.  In vamana's metadata it's called `adjacency_row_index_type` / `adjacency_row_index_datatype` -- so that should be used instead of `px_type` / `px_datatype`.

jparismorgan: Great, thanks, will update to that instead.

lums658 (approved):
This addresses the feedback I had in previous review.

PR #249, opened 2024-02-21
Fix all MSVC compilation issues

dudoslav:
This PR fixes all MSVC compilation issues with the exception of `mmap` issues that are addressed here:
https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/224

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -68,7 +68,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, macos-12, windows-2022]
+        os: [windows-2022]
         # `windows-2022, ` blocked by https://github.com/pybind/pybind11/issues/3445#issuecomment-1525500927
 
     steps:
@@ -81,7 +81,7 @@ jobs:
           brew install automake pkg-config ninja
 
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.16.2
+        uses: pypa/cibuildwheel@v2.16.5
         env:
           CIBW_MANYLINUX_X86_64_IMAGE: "manylinux_2_28"
           CIBW_SKIP: "*-win32 cp27-* cp35-* cp36-* cp37-* pp* *_i686 *musllinux*"
--- a/src/include/detail/graph/vamana.h
+++ b/src/include/detail/graph/vamana.h
@@ -222,7 +222,7 @@ auto greedy_search(
         // @todo (actually it does, but shouldn't need it -- need to
         // investigate) if (result.template insert /*<unique_id>*/ (score, p)) {
         if (result.template insert<unique_id>(score, p)) {
-          q2.template insert /*<unique_id>*/ (score, p);
+          q2.template insert<unique_id>(score, p);
         }
       }
     }
--- a/src/include/test/array_defs.h
+++ b/src/include/test/array_defs.h
@@ -38,6 +38,14 @@
 
 #include "config.h"
 
+namespace {
+
+    std::string operator/(const std::string& lhs, const std::string& rhs) {
+        return (std::filesystem::path{lhs} / rhs).string();
+    }
+
+}
+
 /**
  * @brief  Some default types used in unit tests.
  */
@@ -56,10 +64,10 @@ using test_indices_type = uint64_t;  // Should be same as groundtruth_type
  * so we put it in CMAKE_SOURCE_DIR/external/test_arrays.
  */
 static std::filesystem::path cmake_source_dir{CMAKE_SOURCE_DIR};
-static std::filesystem::path test_data_root{
-    cmake_source_dir.parent_path() / "external" / "test_data"};
-static std::filesystem::path test_array_root{test_data_root / "arrays"};
-static std::filesystem::path test_file_root{test_data_root / "files"};
+static std::string test_data_root =
+    (cmake_source_dir.parent_path() / "external" / "test_data").string();
+static std::string test_array_root{test_data_root / "arrays"};
+static std::string test_file_root{test_data_root / "files"};
 
 /**
  * @brief  Array URIs for arrays used for unit testing of IVF indexes.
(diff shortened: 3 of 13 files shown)


.github/workflows/build-wheels.yml
```diff
@@ -68,7 +68,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, macos-12, windows-2022]
```
dudoslav, line 71: Ignore this file, I'll return it to its original form once all tests pass on my personal fork

src/include/detail/graph/vamana.h
```diff
@@ -222,7 +222,7 @@ auto greedy_search(
         // @todo (actually it does, but shouldn't need it -- need to
         // investigate) if (result.template insert /*<unique_id>*/ (score, p)) {
         if (result.template insert<unique_id>(score, p)) {
-          q2.template insert /*<unique_id>*/ (score, p);
+          q2.template insert<unique_id>(score, p);
```
jparismorgan, line 225: Is this change required? The comment above seems to indicate it makes a difference in something? cc @lums658 if you could explain the comment a bit more?

src/include/test/unit_api_feature_vector.cc
```diff
@@ -110,7 +110,7 @@ TEST_CASE("api: FeatureVector dimension", "[api]") {
   CHECK(dimension(FeatureVector(Vector<int>{1, 2, 3})) == 3);
 }
 
-using TestTypes = std::tuple<float, uint8_t, int32_t, uint32_t, uint64_t>;
+using TestTypes = std::tuple<float, int32_t, uint32_t, uint64_t>;
```
jparismorgan, line 113: What does this change fix?

jparismorgan (commented):
Why are the unit tests not running for this PR? I'm guessing because you're working from a fork instead of a branch in the repo?

src/include/test/unit_api_feature_vector.cc
dudoslav: Good question. `std::uniform_int_distribution` can not use 1 byte size variables in MSVC. The `randomize` function uses this distribution. I don't know how to fix this other way than to delete it.

src/include/detail/graph/vamana.h
dudoslav: Ah, yes. This was a compilation issue for MSVC, saying that it did not expect word `template`.

src/include/test/unit_api_feature_vector.cc
dudoslav: So no `char, uint8_t, int8_t` and friends are allowed.

dudoslav: 1708530796_grim

dudoslav:
> Why are the unit tests not running for this PR? I'm guessing because you're working from a fork instead of a branch in the repo?

Ah, yes, most probably that is why. I'll fix all of your comments and recreate this PR afterwards.

src/include/test/array_defs.h
```diff
@@ -56,10 +64,10 @@ using test_indices_type = uint64_t;  // Should be same as groundtruth_type
  * so we put it in CMAKE_SOURCE_DIR/external/test_arrays.
  */
 static std::filesystem::path cmake_source_dir{CMAKE_SOURCE_DIR};
-static std::filesystem::path test_data_root{
-    cmake_source_dir.parent_path() / "external" / "test_data"};
-static std::filesystem::path test_array_root{test_data_root / "arrays"};
-static std::filesystem::path test_file_root{test_data_root / "files"};
+static std::string test_data_root =
```
lums658, line 67: OK -- this works and is fine with me (replacing `filesystem::path` with `std::string`).

src/include/test/unit_api_feature_vector.cc
lums658: Hum.  Well it has been working.  Is this just an MSVC issue?  If so, address the issue by initializing the vector in some other way -- e.g., fill it random numbers generated in some other way, or with the digits of pi, or something else when in MSVC and the type is `uint8_t`..  `uint8_t` is an important type for vector search so we need to be able to test reading and writing it.

In general, making tests pass (even compilation) by removing the offending instance from the test suite is very bad practice.

In fact, I would claim this case did exactly what it is supposed to do -- it found a problem -- even if happened at compilation time rather than run time.  Again, when a unit test fails, it is telling you to fix the code, not the test.

Unit tests exist to tell you when something has gone amiss when you change your code (or your compilation environment).

(And good catch, Paris.)

src/include/utils/filesystem.h
```diff
@@ -0,0 +1,48 @@
+/**
```
lums658, line 1: Do we need this any longer?  If not, it should be deleted.

src/include/detail/graph/vamana.h
lums658: Oh this is one of the issues that has been impacting performance -- whether or not to explicitly maintain unique keys in the heap.  Better in this case to take `template` out rather than put `unique_id` back in: 
```c++
         if (result.template insert<unique_id>(score, p)) {
           // q2.template insert <unique_id>(score, p);
           q2.insert(score, p);
```

src/include/test/unit_utils.cc
```diff
@@ -33,6 +33,14 @@
 #include "config.h"
 #include "utils/utils.h"
 
+namespace {
```
lums658, line 36: Do we need this any longer?  If we are still using `stdx::filesystem::path`, then maybe it should go into that header?  I think with the approach we are taking now, we shouldn't need `stdx::filesystem::path`?

lums658 (changes requested):
The change from `std::filesystem` to `std::string` seems to be the right way to go and the changes in this PR in that regard look fine.

A few smallish changes to be addressed elsewhere:
* Undo the change to the unit test (i.e., put back `uint8_t` into the type list)
* Remove `stdx::filesystem::path` and associated header files, as I don't think we are using that any longer?
* Make a different change to the `template` / `unique_id` change

src/include/test/unit_utils.cc
dudoslav: I did commit this by mistake, will remove

PR #250, opened 2024-02-23
Fix all MSVC compilation issues

dudoslav:
This PR is reopened and merged from main PR: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/249

Comments from that PR:

[ ]    Undo the change to the unit test (i.e., put back uint8_t into the type list)
[ ]    Remove stdx::filesystem::path and associated header files, as I don't think we are using that any longer?
[ ]    Make a different change to the template / unique_id change

Note: The previous PR got closed since the merged to branch got deleted (merged into main)


dudoslav:
It seems that the windows tests timeout for some reason:
```
        Start  4: unit_adj_list_test
   4/43 Test  #4: unit_adj_list_test ...................   Passed    0.68 sec
        Start  5: unit_algorithm_test
   5/43 Test  #5: unit_algorithm_test ..................***Timeout 1500.02 sec
```

I will investigate further. https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/8017844099/job/21902529398?pr=250

src/include/test/test_utils.h
```diff
@@ -37,24 +37,34 @@
...
     for (auto& x : r) {
       x = dist(gen);
     }
-    return;
   } else {
-    std::uniform_int_distribution<std::ranges::range_value_t<R>> dist(
-        std::get<0>(range), std::get<1>(range));
-    for (auto& x : r) {
-      x = dist(gen);
+    if constexpr (sizeof(element_type) == 1u) {
```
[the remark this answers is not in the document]
lums658: I would suggest using `uint16_t` or `int16_t` so that you can be sure of the size.  I am also wondering if `%` is better than `/`.  (I don't have an answer for sure -- for some prngs `/` may be better statistically -- but that probably isn't the case for modern libraries.)  But using % would also let you guarantee that you have an 8 bit thing at the end.  Or you could just mask off 8 bits.

src/include/test/unit_utils.cc
```diff
@@ -33,6 +33,14 @@
 #include "config.h"
 #include "utils/utils.h"
 
+namespace {
```
lums658, line 36: Ah -- interesting -- I see how this is useful now.

lums658 (approved):
This addresses all the comments I made in the previous PR.

src/include/test/test_utils.h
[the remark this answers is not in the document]
dudoslav: I agree with using `uint16_t` instead of `short`. Regarding the second part of your commend, using `/` or `%`. The modulo operator should work fine, but in that case we can simply just downcast the value right? (modulo 256 is the same as casting to `uint8_t`). Do you want me to change it? :)

src/include/test/unit_utils.cc
dudoslav: Yea, but it still isn't good :/, since this operator is defined in header it pollutes the top level namespace. It works, I just don't like it, I would prefer if it was only defined in compilation unit.

src/cmake/Superbuild.cmake
```diff
@@ -150,9 +150,9 @@ add_custom_target(
 
 # make check
 add_custom_target(check
-  COMMAND ${CMAKE_CTEST_COMMAND} --test-dir ${CMAKE_CURRENT_BINARY_DIR}/libtiledbvectorsearch --output-on-failure
+  COMMAND ${CMAKE_CTEST_COMMAND} -C $<CONFIG> --test-dir ${CMAKE_CURRENT_BINARY_DIR}/libtiledbvectorsearch --output-on-failure
 )
 
 add_custom_target(check-ci
-  COMMAND ${CMAKE_CTEST_COMMAND} --test-dir ${CMAKE_CURRENT_BINARY_DIR}/libtiledbvectorsearch --output-on-failure -E \"unit_slicing|unit_ivf_index\"
+  COMMAND ${CMAKE_CTEST_COMMAND} -C $<CONFIG> --test-dir ${CMAKE_CURRENT_BINARY_DIR}/libtiledbvectorsearch --output-on-failure --extra-verbose -E \"unit_slicing|unit_ivf_index|unit_logging\"
```
dudoslav, line 157: For some reason `unit_logging` sleep interval is so imprecise that it fails the tests. I don't know how to handle this issue and if it is severe enough to care about:

https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/8064265455/job/22027798473?pr=250

src/include/test/unit_tdb_io.cc
```diff
@@ -184,10 +189,14 @@ TEST_CASE("tdb_io: create group", "[tdb_io]") {
   auto num_members = read_group.member_count();
   CHECK(num_members == 1);
   auto a = read_group.member(0);
-  CHECK(a.uri() == "file:///tmp/tmp_group/ids");
+  // CHECK(
```
dudoslav, line 192: I am not 100% sure what these tests are supposed to test but our implementation of VFS seems to not be aware of Windows paths and thus prepends global paths with `/` so paths like `C:\Users` gets converted to `/C:\Users` and that breaks this comparison.

src/include/utils/utils.h
```diff
@@ -64,7 +64,7 @@ std::string get_filename(const std::string& filename) {
 }
 
 bool local_directory_exists(const std::string& path) {
-  std::filesystem::path directoryPath(path);
+  std::string directoryPath(path);
```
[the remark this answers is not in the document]
dudoslav: Resolved

jparismorgan (approved):

PR #251, opened 2024-02-23
In `tdb_matrix_with_ids` fix data loading, the default type of ids, and the attribute check when loading data

jparismorgan:
### What
When adding IDs to `feature_vector_array.h` I found three bugs in `tdb_matrix_with_ids`:
1. We were loading data from a column major matrix (two dimensions) instead of a row major vector of IDs (one dimension). Here we fix that and also fix `fill_and_write_matrix()` to correctly create both a matrix at `uri` and a vector of IDs at `ids_uri`.
2. Update the default type of IDs to `uint64_t`
3. Fix the check for whether the data in `ids_uri` is the same type as the template parameter in the class.

### Testing
* Added unit tests in this PR
* My `feature_vector_array.h` changes now also pass unit tests

### DIFF
--- a/src/include/detail/linalg/tdb_matrix_with_ids.h
+++ b/src/include/detail/linalg/tdb_matrix_with_ids.h
@@ -50,7 +50,7 @@
  */
 template <
     class T,
-    class IdsType = size_t,
+    class IdsType = uint64_t,
     class LayoutPolicy = stdx::layout_right,
     class I = size_t>
 class tdbBlockedMatrixWithIds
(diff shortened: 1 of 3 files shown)


src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -50,7 +50,7 @@
  */
 template <
     class T,
-    class IdsType = size_t,
+    class IdsType = uint64_t,
```
lums658, line 53: Yep.  I think the rule of thumb should be -- don't use size_t for data -- only for indexing into data (which is how it is defined and intended to be used).  I have a bad habit of over-using it.

src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -188,11 +188,12 @@ class tdbBlockedMatrixWithIds
 
     std::string attr_name = attr.name();
     tiledb_datatype_t attr_type = attr.type();
-    if (attr_type != tiledb::impl::type_to_tiledb<T>::tiledb_type) {
+    if (attr_type != tiledb::impl::type_to_tiledb<IdsType>::tiledb_type) {
```
lums658, line 191: I generally have a practice of using an indirection for types used within a class / function template, rather than the function template name.  Especially for classes, this lets you have uniform identifier names, even when you don't have templated types.  Hence, `using feature_type = T; using ids_type = IdsType` and so on.  It also localizes what you would have to change if you decided to do something different to determine what IdsType is.  Just a suggestion.

lums658 (approved):
Looks good to me.

src/include/detail/linalg/tdb_matrix_with_ids.h
jparismorgan: Sure, updated.

PR #253, opened 2024-02-27
Optimize query latency for Index and ObjectIndex classes

NikolaosPapailiou:
This adds the following query latency optimizations:
- `ObjectIndex`: 
   - Open `object_metadata_array` in the constructor and reuse the opened array in all queries.
   - Add option to load the `object_metadata` in memory during the constructor. This allows metadata filtering and retrieval directly from main memory without reading from the TileDB array.
- `Index`: Avoid checking existence of updates array and `has_updates` metadata value in the critical path of queries.

### DIFF
--- a/apis/python/src/tiledb/vector_search/object_api/object_index.py
+++ b/apis/python/src/tiledb/vector_search/object_api/object_index.py
@@ -1,8 +1,7 @@
-import base64
+import json
 from collections import OrderedDict
 from typing import Any, Dict, List, Mapping, Optional
 
-import json
 import numpy as np
 
 import tiledb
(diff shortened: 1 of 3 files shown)


apis/python/src/tiledb/vector_search/object_api/object_index.py
```diff
@@ -29,12 +28,14 @@ def __init__(
...
         timestamp=None,
         load_embedding: bool = True,
+        load_metadata_in_memory: bool = True,
         **kwargs,
     ):
         with tiledb.scope_ctx(ctx_or_config=config):
             self.uri = uri
             self.config = config
             self.timestamp = timestamp
+            self.load_metadata_in_memory = load_metadata_in_memory
```
jparismorgan, line 38: (Non-blocking) What is the case where we will want to set `load_metadata_in_memory` to false? Is it about the memory pressure / speed? Curious if we should just always do this and not expose an option.

NikolaosPapailiou: For large datasets, for which the dataset can't fit in memory or we don't want to impose large memory requirements to the client, we want to have this processing happen from the TileDB array in an out-of-core manner. This is similar to how we perform vector search queries, we offer both memory based and out-of-core implementations.

jparismorgan: Makes sense, thanks.

PR #254, opened 2024-02-29
Add support for empty and matrices and vectors

jparismorgan:
### What
When we add type-erased APIs we will have a flow where we need to create an empty index:
```
def create(
    uri: str,
    dimensions: int,
    vector_type: np.dtype,
    id_type: np.dtype = np.uint32,
    adjacency_row_index_type: np.dtype = np.uint32,
    group_exists: bool = False,
    storage_version: str = STORAGE_VERSION,
    **kwargs,
) -> VamanaIndex:
      ctx = vspy.Ctx(config)
      index = vspy.IndexVamana(
          feature_type=np.dtype(vector_type).name, 
          id_type=np.dtype(id_type).name, 
          adjacency_row_index_type=np.dtype(adjacency_row_index_type).name, 
          dimension=dimensions,
        )
      index.write_index(ctx, uri)
      return VamanaIndex(uri=uri, config=config, memory_budget=1000000) # this is a Python VamanaIndex object.
```
When we do that we need to be able to write empty arrays and vectors. This writing of empty arrays and vectors works, but when we try to read them back we use code like this for both vectors and matrices to fill in the returned data:
```
if (start_pos == 0) {
    start_pos = array_rows_.template domain<domain_type>().first;
  }
  if (end_pos == 0) {
    end_pos = array_rows_.template domain<domain_type>().second + 1;
  }
```
Because of this we then try to run the query:
```
tiledb::Query query(ctx, *array_);
  query.set_subarray(subarray).set_data_buffer(
      attr_name, data_.data(), vec_rows_);
  tiledb_helpers::submit_query(tdb_func__, uri, query);
```
But we crash because `data_.data()` is actually empty and so there is no where to write the data to.

The fix that this PR makes is change the default behavior to not update the start and end positions and to return early if there is nothing to read. Ideally we would get to a point where we never need the default behavior and the caller is responsible for correctly asking for how much data to read. This is based off of what Python does where it has the caller manage how much to read, versus the reader automatically reading all it can. Though please do let me know if you have a different suggestion on how to fix this.

### Testing
* Added new unit tests.
* Existing unit tests pass.

### DIFF
--- a/apis/python/src/tiledb/vector_search/module.cc
+++ b/apis/python/src/tiledb/vector_search/module.cc
@@ -640,7 +640,8 @@ PYBIND11_MODULE(_tiledbvspy, m) {
          size_t start_pos,
          size_t end_pos,
          uint64_t timestamp) -> std::vector<uint32_t> {
-        auto r = read_vector<uint32_t>(ctx, uri, start_pos, end_pos, timestamp);
+        auto r = read_vector<uint32_t>(
+            ctx, uri, start_pos, end_pos, timestamp, true);
         return r;
       });
   m.def(
@@ -650,7 +651,8 @@ PYBIND11_MODULE(_tiledbvspy, m) {
          size_t start_pos,
          size_t end_pos,
          uint64_t timestamp) -> std::vector<uint64_t> {
-        auto r = read_vector<uint64_t>(ctx, uri, start_pos, end_pos, timestamp);
+        auto r = read_vector<uint64_t>(
+            ctx, uri, start_pos, end_pos, timestamp, true);
         return r;
       });
 
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -65,9 +65,9 @@ void create_empty_for_matrix(
   tiledb::Domain domain(ctx);
   domain
       .add_dimension(tiledb::Dimension::create<int>(
-          ctx, "rows", {{0, std::max(0, (int)rows - 1)}}, row_extent))
+          ctx, "rows", {{0, (int)rows - 1}}, row_extent))
       .add_dimension(tiledb::Dimension::create<int>(
-          ctx, "cols", {{0, std::max(0, (int)cols - 1)}}, col_extent));
+          ctx, "cols", {{0, (int)cols - 1}}, col_extent));
 
   tiledb::ArraySchema schema(ctx, TILEDB_DENSE);
 
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -57,8 +57,8 @@ template <class T, class LayoutPolicy = stdx::layout_right, class I = size_t>
 void create_empty_for_matrix(
     const tiledb::Context& ctx,
     const std::string& uri,
-    size_t rows,  // number of dimensions
-    size_t cols,  // number of vectors
+    size_t rows,
+    size_t cols,
     size_t row_extent,
     size_t col_extent,
     std::optional<tiledb_filter_type_t> filter = std::nullopt) {
@@ -143,9 +143,9 @@ void write_matrix(
 
   std::vector<int32_t> subarray_vals{
       0,
-      std::max((int)A.num_rows() - 1, 0),
+      (int)A.num_rows() - 1,
       (int)start_pos,
-      std::max((int)start_pos + (int)A.num_cols() - 1, 0)};
+      (int)start_pos + (int)A.num_cols() - 1};
 
   // Open array for writing
   auto array = tiledb_helpers::open_array(
@@ -190,7 +190,7 @@ void create_empty_for_vector(
     std::optional<tiledb_filter_type_t> filter = std::nullopt) {
   tiledb::Domain domain(ctx);
   domain.add_dimension(tiledb::Dimension::create<int>(
-      ctx, "rows", {{0, std::max(0, (int)rows - 1)}}, row_extent));
+      ctx, "rows", {{0, (int)rows - 1}}, row_extent));
 
   // The array will be dense.
   tiledb::ArraySchema schema(ctx, TILEDB_DENSE);
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -57,17 +57,17 @@ template <class T, class LayoutPolicy = stdx::layout_right, class I = size_t>
 void create_empty_for_matrix(
     const tiledb::Context& ctx,
     const std::string& uri,
-    size_t rows,
-    size_t cols,
+    size_t rows,  // number of dimensions
+    size_t cols,  // number of vectors
     size_t row_extent,
     size_t col_extent,
     std::optional<tiledb_filter_type_t> filter = std::nullopt) {
   tiledb::Domain domain(ctx);
   domain
       .add_dimension(tiledb::Dimension::create<int>(
-          ctx, "rows", {{0, (int)rows - 1}}, row_extent))
+          ctx, "rows", {{0, std::max(0, (int)rows - 1)}}, row_extent))
       .add_dimension(tiledb::Dimension::create<int>(
-          ctx, "cols", {{0, (int)cols - 1}}, col_extent));
+          ctx, "cols", {{0, std::max(0, (int)cols - 1)}}, col_extent));
 
   tiledb::ArraySchema schema(ctx, TILEDB_DENSE);
 
@@ -143,9 +143,9 @@ void write_matrix(
 
   std::vector<int32_t> subarray_vals{
       0,
-      (int)A.num_rows() - 1,
+      std::max((int)A.num_rows() - 1, 0),
       (int)start_pos,
-      (int)start_pos + (int)A.num_cols() - 1};
+      std::max((int)start_pos + (int)A.num_cols() - 1, 0)};
 
   // Open array for writing
   auto array = tiledb_helpers::open_array(
@@ -190,7 +190,7 @@ void create_empty_for_vector(
     std::optional<tiledb_filter_type_t> filter = std::nullopt) {
   tiledb::Domain domain(ctx);
   domain.add_dimension(tiledb::Dimension::create<int>(
-      ctx, "rows", {{0, (int)rows - 1}}, row_extent));
+      ctx, "rows", {{0, std::max(0, (int)rows - 1)}}, row_extent));
 
   // The array will be dense.
   tiledb::ArraySchema schema(ctx, TILEDB_DENSE);
@@ -289,7 +289,8 @@ std::vector<T> read_vector(
     const std::string& uri,
     size_t start_pos,
     size_t end_pos,
-    size_t timestamp = 0) {
+    size_t timestamp = 0,
+    bool read_full_vector = false) {
   scoped_timer _{tdb_func__ + " " + std::string{uri}};
 
   tiledb::TemporalPolicy temporal_policy =
@@ -308,14 +309,20 @@ std::vector<T> read_vector(
   auto dim_num_{domain_.ndim()};
   auto array_rows_{domain_.dimension(0)};
 
-  if (start_pos == 0) {
-    start_pos = array_rows_.template domain<domain_type>().first;
-  }
-  if (end_pos == 0) {
-    end_pos = array_rows_.template domain<domain_type>().second + 1;
+  if (read_full_vector) {
+    if (start_pos == 0) {
+      start_pos = array_rows_.template domain<domain_type>().first;
+    }
+    if (end_pos == 0) {
+      end_pos = array_rows_.template domain<domain_type>().second + 1;
+    }
   }
 
   auto vec_rows_{end_pos - start_pos};
+
+  if (vec_rows_ == 0) {
+    return {};
+  }
 
   auto attr_num{schema_.attribute_num()};
   auto attr = schema_.attribute(idx);
@@ -350,7 +357,7 @@ std::vector<T> read_vector(
 template <class T>
 std::vector<T> read_vector(
     const tiledb::Context& ctx, const std::string& uri, size_t timestamp = 0) {
-  return read_vector<T>(ctx, uri, 0, 0, timestamp);
+  return read_vector<T>(ctx, uri, 0, 0, timestamp, true);
 }
 
 template <class T>
--- a/src/include/detail/linalg/tdb_matrix.h
+++ b/src/include/detail/linalg/tdb_matrix.h
@@ -130,7 +130,7 @@ class tdbBlockedMatrix : public MatrixBase {
    */
   tdbBlockedMatrix(const tiledb::Context& ctx, const std::string& uri) noexcept
     requires(std::is_same_v<LayoutPolicy, stdx::layout_left>)
-      : tdbBlockedMatrix(ctx, uri, 0, 0, 0, 0, 0, 0) {
+      : tdbBlockedMatrix(ctx, uri, 0, 0, 0, 0, 0, 0, true) {
   }
 
   /**
@@ -150,7 +150,7 @@ class tdbBlockedMatrix : public MatrixBase {
       size_t upper_bound,
       size_t timestamp = 0)
     requires(std::is_same_v<LayoutPolicy, stdx::layout_left>)
-      : tdbBlockedMatrix(ctx, uri, 0, 0, 0, 0, upper_bound, timestamp) {
+      : tdbBlockedMatrix(ctx, uri, 0, 0, 0, 0, upper_bound, timestamp, true) {
   }
 
   /** General constructor */
@@ -162,7 +162,8 @@ class tdbBlockedMatrix : public MatrixBase {
       size_t first_col,
       size_t last_col,
       size_t upper_bound,
-      size_t timestamp)
+      size_t timestamp,
+      bool read_full_matrix)
     requires(std::is_same_v<LayoutPolicy, stdx::layout_left>)
       : tdbBlockedMatrix(
             ctx,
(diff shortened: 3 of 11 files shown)


apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -425,7 +426,8 @@ static void declarePartitionedMatrix(
           const std::string&,
           const std::string&,                // id_uri
           const std::vector<Indices_Type>&,  // partition list to load
-          size_t>(),                         // upper_bound
+          size_t,
+          bool>(),  // upper_bound
```
NikolaosPapailiou, line 430: nit: `size_t, // upper_bound`

src/include/detail/linalg/tdb_io.h
```diff
@@ -308,15 +309,21 @@ std::vector<T> read_vector(
   auto dim_num_{domain_.ndim()};
   auto array_rows_{domain_.dimension(0)};
 
-  if (start_pos == 0) {
-    start_pos = array_rows_.template domain<domain_type>().first;
-  }
-  if (end_pos == 0) {
-    end_pos = array_rows_.template domain<domain_type>().second + 1;
+  if (read_full_vector) {
```
NikolaosPapailiou, line 312: The API of this function seems a bit inconsistent to me. We already have the notion of if `start_pos, end_pos == 0` we read full vector based on domains. Why do we need an extra `read_full_vector` argument vector for?

The client can either use 0 for all positions or define specific slice to read.

We either need to use domain for size and pass `start_pos, end_pos == 0` or use max domain and maintain the size on the client. Not sure what this will give us on top of that.

jparismorgan: > We already have the notion of if `start_pos, end_pos == 0` we read full vector based on domains. Why do we need an extra read_full_vector argument vector for?

We need some way of telling this function not to read the full vector. Right now if you pass `start_pos, end_pos == 0` then we read the full vector, but if there is no where to read the vector into, i.e. in this case, then we crash:
```
tiledb::Query query(ctx, *array_);
  query.set_subarray(subarray).set_data_buffer(
      attr_name, data_.data(), vec_rows_);
  tiledb_helpers::submit_query(tdb_func__, uri, query);
```
Does that make sense? This is basically a flag to say "`start_pos, end_pos == 0` means read nothing vs means read everything".

NikolaosPapailiou: Could we have 2 functions?
- with start, end pos
- with just read from domain

jparismorgan: So there are already two functions, one where you don't pass positions:
```
template <class T>
std::vector<T> read_vector(
    const tiledb::Context& ctx, const std::string& uri, size_t timestamp = 0) {
  return read_vector<T>(ctx, uri, 0, 0, timestamp, true);
}
```
and one where do you:
```
template <class T>
std::vector<T> read_vector(
    const tiledb::Context& ctx,
    const std::string& uri,
    size_t start_pos,
    size_t end_pos,
    size_t timestamp = 0,
    bool read_full_vector = false) {
```
The issue I was seeing is that some callers call into the one with positions with 0 values, and if I don't keep this logic inside the one with positions then we fail tests. But I'l investigate further and try to fix those because you're right, having two separate functions is much simpler.

jparismorgan: Okay updated to your suggestion, now we just have two functions.

apis/python/src/tiledb/vector_search/module.py
```diff
@@ -36,15 +36,15 @@ def load_as_matrix(
     a = tiledb.ArraySchema.load(path, ctx=tiledb.Ctx(config))
     dtype = a.attr(0).dtype
     if dtype == np.float32:
-        m = tdbColMajorMatrix_f32(ctx, path, 0, 0, 0, size, 0, timestamp)
+        m = tdbColMajorMatrix_f32(ctx, path, 0, 0, 0, size, 0, timestamp, True)
```
NikolaosPapailiou, line 39: Let avoid changing this if not needed

NikolaosPapailiou (approved):

apis/python/src/tiledb/vector_search/module.py
jparismorgan: So I did this because I didn't want to change the behavior of the existing Python implementation at all, and this is the flag which keeps the old implementation. It would be cleaner looking to set the default value for this argument (`bool read_full_matrix` in `src/include/detail/linalg/tdb_matrix.h`) to `True`, but I didn't set a default value because I'd like to migrate to not needing `bool read_full_matrix` at all (and instead having callers pass in the size to read correctly), and in the meantime I want it to be very explicit whether we are doing `read_full_matrix` or not.

lums658, line 39: I don't love this interface to `tdbMatrix`.  I think what I would rather see (and I think have done elsewhere for something else) is to use an overload so that if we want to read the whole thing, we don't pass in a limit at all.  

(Using special values of variables can be bad if the special value is actually a valid value.)

src/include/detail/linalg/tdb_io.h
```diff
@@ -43,6 +43,77 @@
...
+namespace {
+
+template <class T>
+std::vector<T> read_vector_helper(
+    const tiledb::Context& ctx,
+    const std::string& uri,
+    size_t start_pos,
+    size_t end_pos,
+    size_t timestamp,
+    bool read_full_vector) {
```
lums658, line 55: Rather than having a flag, we should just omit `start_pos` and `end_pos` if we want to read the whole thing.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -130,7 +130,7 @@ class tdbBlockedMatrix : public MatrixBase {
    */
   tdbBlockedMatrix(const tiledb::Context& ctx, const std::string& uri) noexcept
     requires(std::is_same_v<LayoutPolicy, stdx::layout_left>)
-      : tdbBlockedMatrix(ctx, uri, 0, 0, 0, 0, 0, 0) {
+      : tdbBlockedMatrix(ctx, uri, 0, 0, 0, 0, 0, 0, true) {
```
lums658, line 133: Again, we should not use explicit flag, but just omit all of the view parameters.

And, if there are issues related to ambiguity because of the suggested overloading (I don't think there will be), we should use a class that is a start-stop pair, rather than separate integers.  This is something I have had in mind to do anyway.

lums658 (changes requested):
Main comment -- Instead of having a flag to indicate whether to read the entire array (i.e., use extents as dimensions to read into), use an overload that simply omits any specified limits.

Using special values of parameters to indicate different cases is probably not a good practice in general, unless the special value is absolutely, positively not a valid value (such as a `nullptr`).  Mea culpa.

src/include/detail/linalg/tdb_io.h
jparismorgan: So this is just a helper function in an anonymous namespace, the actual user-facing APIs are:
```
template <class T>
std::vector<T> read_vector(
    const tiledb::Context& ctx,
    const std::string& uri,
    size_t start_pos,
    size_t end_pos,
    size_t timestamp = 0) {
  return read_vector_helper<T>(ctx, uri, start_pos, end_pos, timestamp, false);

template <class T>
std::vector<T> read_vector(
    const tiledb::Context& ctx, const std::string& uri, size_t timestamp = 0) {
  return read_vector_helper<T>(ctx, uri, 0, 0, timestamp, true);
}
```
Do you still want to change it? This approach doesn't seem that bad to me.

apis/python/src/tiledb/vector_search/module.py
jparismorgan: Changed in the update so will resolve this and re-request your review!

jparismorgan: Sure thanks, updated the constructors so that callers can use the appropriate one without this flag.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: Sure, thanks, done. Will keep that in mind as well.

jparismorgan:
Note that the Windows build is failing but I will fix before landing. Just opening up this PR again for review so that I can catch @lums658 today while he's still online.

src/include/detail/linalg/tdb_io.h
```diff
@@ -43,6 +43,77 @@
...
+  using domain_type = int32_t;
+  const size_t idx = 0;
+
+  auto domain_{schema_.domain()};
+
+  auto dim_num_{domain_.ndim()};
+  auto array_rows_{domain_.dimension(0)};
+
+  if (read_full_vector) {
+    if (start_pos == 0) {
```
lums658, line 75: In thinking about this, just like `stop_pos == 0` should have no special meaning as a flag, `stop_pos == 0` should have no special meaning (0 should not be a sentinel).  That is, we might want to read a "half-full" vector -- starting at a given place and reading to the end or starting at the beginning and reading to a given place -- and the "given place" in either case could be zero, which would be the wrong thing since right now it is either both or neither.  We can do that currently (though there is a bug -- if you want to actually start at zero but the array doesn't start at zero). So we shouldn't be using a sentinel value for either `start` or `stop`

Unfortunately, we can't determine which half-full case we want with an overload that has one of `start` or `stop` omitted.

There are four cases:
- array begin to array end -- `read_full_vector == true` in the helper -- at which point `start` and `stop` are sentinel values and "end" is indicating by the overloaded function passing the sentinel values to the helper function.
- array begin to `stop` -- not supported in current PR (previously was)
- `start` to array end -- not supported in current PR (previously was, though a bug for `start == 0`)
- `start` to `stop` -- in the current PR, 0 is a valid value for either `start` or `stop`
where zero is a valid value for `start` or `stop` and array begin or array end (or both) may be zero or may not be zero.  Something like Python's range notation would be really helpful....

While we are waiting for C++ to introduce that notation, we need a way to specify "array begin" or "array end" for either `start` or `stop` -- without using a sentinel.  Or at least not using a sentinel that is a value likely to actually be used.  
* One approach would be to use a sentinel that is unlikely to ever be used, like -MAXINT and MAXINT.  Actually, in that case, collisions would be benign -- meaning "read from the beginning" and "read from this value" are the same.
* Or, we could use `std::optional` -- an empty value would mean from the beginning/end for `start`/`stop`.  The notation from the caller side is kind of indicative of what is desired.
```c++
template <class T> std::vector<T> read_vector(const tiledb::Context& ctx, const std::string& uri, 
std::optional<int32_t> start_pos, std::optional<int32_t> end_pos, std::optional<size_t> timestamp);

auto foo = read_vector(ctx, "a.tdb", 0, {}, {});
auto bar = read_vector(ctx, "b.tdb", {}, {}, {}); // Could still have overload: read_vector(ctx, "b.tdb", {});
```
(Note that I am also suggesting we use optional for timestamp, though maybe default value of zero really is what is intended.)

Note also, and very important, `start` and `stop` should be `int32_t` since that is what the vectors and matrix types assume (we should use an appropriate type alias so that they are all consistent).

The constructors for `tdbMatrix` (et al) have the same issues (but twice as many variables for factorially more combinations).

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -114,6 +114,26 @@ class tdbBlockedMatrix : public MatrixBase {
   // size_t pending_row_offset{0};
   // size_t pending_col_offset{0};
 
+  static size_t get_last_row(
```
lums658, line 117: Hmm.  I don't think we want to open and close the array to get this info, only to open the array again.

However, I think this is a moot point, per the extensive comment above, as I think we should move everything into one constructor and use `std::optional` to indicate whether or not we want begin/end of the array.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -122,21 +142,39 @@ class tdbBlockedMatrix : public MatrixBase {
...
    *
    * @param ctx The TileDB context to use.
    * @param uri URI of the TileDB array to read.
+   * @param upper_bound The maximum number of vectors to read. Set to 0 for no
+   * upper bound.
+   * @param temporal_policy The TemporalPolicy to use for reading the array
+   * data.
    */
-  tdbBlockedMatrix(const tiledb::Context& ctx, const std::string& uri) noexcept
+  tdbBlockedMatrix(
```
lums658, line 156: Per the extensive comment above, I think the correct interface is to have just a single constructor with `std::optional` for `start_row`, et al.  This will simplify things a lot (code-wise and conceptually) and also allow the array to just be opened once in order to access its schema (it is opened in the initializer of the general constructor) so that `begin_row`, `last_row`, `begin_col`, and `last_col` are cheaply obtained.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -147,10 +185,39 @@ class tdbBlockedMatrix : public MatrixBase {
   tdbBlockedMatrix(
       const tiledb::Context& ctx,
       const std::string& uri,
+      size_t first_col,
```
lums658, line 188: Per the extensive comment above, `first_col` et al should be the same type as the domain -- we should probably have a global type alias for this.  In `tdbMatrix` we use `row_domain_type` and `col_domain_type` (both of which are `int32_t` -- but we should not hard code that everywhere).  We could just use those:
```
tdbBlockedMatrix(
       const tiledb::Context& ctx,
       const std::string& uri,
       std::optional<col_domain_type> first_row, // etc
```

lums658 (changes requested):
There is a fundamental flaw in the way we have proposed not to have sentinel values for `start` and `stop` (et al).  Basically, we need to be able to specify whether to ignore or use *any* of `start` / `stop` quantities when reading vectors or matrices -- and that isn't possible to do with overloading.  

There is also the issue of having to open and close an array to get its dimensions, before opening the array (again getting its schema) and getting its data.

The only solution I can see to this is to use `std::optional` to indicate whether or not to use the `start` / `stop` value, or whether to use the begin / end of the array.  (Presence or absence of a value is what `std::optional` is for, after all.). This approach also expresses intent of the caller -- using `{}` instead of a value says "no value" -- i.e., in this case, use the appropriate limit.  

Another important change, in conjunction with using `std::optional` is to use the same type as the domain, since we are representing values in that domain.  `tdbMatrix` has some type aliases that can be used for this, or we could define something globally.

These changes should simplify quite a few things.

jparismorgan:
> There is a fundamental flaw in the way we have proposed not to have sentinel values for start and stop (et al). Basically, we need to be able to specify whether to ignore or use any of start / stop quantities when reading vectors or matrices -- and that isn't possible to do with overloading.

Yes. That is why I went with my original approach of using a flag before you requested: `Instead of having a flag to indicate whether to read the entire array (i.e., use extents as dimensions to read into), use an overload that simply omits any specified limits`. I will change back but use `std::optional` this time.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -351,62 +405,75 @@ class tdbPreLoadMatrix : public tdbBlockedMatrix<T, LayoutPolicy, I> {
...
+   * all vectors.
+   */
   tdbPreLoadMatrix(
       const tiledb::Context& ctx,
       const std::string& uri,
       size_t num_array_rows,
       size_t num_array_cols,
       size_t upper_bound,
-      const tiledb::TemporalPolicy& temporal_policy)
+      uint64_t timestamp)
```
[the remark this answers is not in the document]
jparismorgan: I thought we decided with @NikolaosPapailiou that we should be using `TemporalPolicy` everywhere?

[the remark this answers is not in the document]
lums658: That is what I thought too.  But isn’t the above changing it from `TemporalPolicy` to use `timestamp`?

jparismorgan:
To make the next review easier I'm going to close this PR and re-open as a separate PR for reading a vector and one for reading `tdb_matrix`.

src/include/detail/linalg/tdb_matrix.h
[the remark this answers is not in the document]
jparismorgan: Resolving because not relevant anymore with the new design.

apis/python/src/tiledb/vector_search/module.py
```diff
@@ -35,20 +35,38 @@ def load_as_matrix(
...
-    elif dtype == np.int32:
-        m = tdbColMajorMatrix_i64(ctx, path, 0, 0, 0, size, 0, timestamp)
-    elif dtype == np.uint8:
-        m = tdbColMajorMatrix_u8(ctx, path, 0, 0, 0, size, 0, timestamp)
-    # elif dtype == np.uint64:
-    #     return tdbColMajorMatrix_u64(ctx, path, size, timestamp)
+    if size == 0:
+        # Read all rows and all cols. Set no upper_bound.
+        if dtype == np.float32:
+            m = tdbColMajorMatrix_f32(ctx, path, 0, timestamp)
```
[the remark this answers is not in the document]
jparismorgan: Sure, let's leave it then and change if we see the need to.

src/include/detail/linalg/tdb_io.h
jparismorgan: Okay sounds good, moved this code over to a new PR as-is.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: Sure, updated to new approach.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: Sure, updated to new approach.

src/include/detail/linalg/tdb_matrix.h
jparismorgan: Sure, created a new story for this as: https://app.shortcut.com/tiledb-inc/story/42696

src/include/detail/linalg/tdb_io.h
jparismorgan: Agree there is certainly more we can improve here, but I would propose to go with the current approach because it is working and is simple. If we find scenarios where it doesn't support our needs, then we can update it then.

PR #255, opened 2024-02-29
Release 0.0.22 against TileDB-Py 0.26 / TileDB 2.19

ihnorton:
We need a release against TileDB 2.19 with the newer APIs, pending other updates for 2.20.


NikolaosPapailiou (approved):

dudoslav (approved):
LGTM, I assume that all downloads in CMAKE can stay as recent as possible.

PR #257, opened 2024-03-01
Add support to execute ObjectApi ingestion entirely using TaskGraphs

NikolaosPapailiou:
This allows the cloud execution of the ingestion driver in a separate TaskGraph enabling:
- Triggering jobs without having access permissions on the client (setting `access_credential_name`)
- Triggering jobs without having all the required Python dependencies for readers and embedding functions.

Tested in Cloud
- [Driver TaskGraph](https://cloud.tiledb.com/compute/logs/taskgraphs/TileDB-Inc/a3f03968-d0fe-491b-afff-80dcb336a769)
- [Embedding generation TaskGraph](https://cloud.tiledb.com/compute/logs/taskgraphs/TileDB-Inc/6ae250e9-0353-442a-9ec2-b93278cb160e)

### DIFF
--- a/apis/python/src/tiledb/vector_search/object_api/embeddings_ingestion.py
+++ b/apis/python/src/tiledb/vector_search/object_api/embeddings_ingestion.py
@@ -49,7 +49,7 @@ def ingest_embeddings_with_driver(
     ):
         import tiledb
 
-        def install_extra_modules():
+        def install_extra_worker_modules():
             if extra_driver_modules is not None:
                 import os
                 import subprocess
@@ -63,7 +63,7 @@ def ingest_embeddings_with_driver(
                         [sys.executable, "-m", "pip", "install", module]
                     )
 
-        install_extra_modules()
+        install_extra_worker_modules()
         import importlib
 
         importlib.reload(tiledb)
@@ -134,7 +134,7 @@ def ingest_embeddings_with_driver(
             config: Optional[Mapping[str, Any]] = None,
             extra_worker_modules: Optional[List[str]] = None,
         ):
-            def install_extra_modules():
+            def install_extra_driver_modules():
                 if extra_worker_modules is not None:
                     import os
                     import subprocess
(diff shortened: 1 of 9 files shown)


apis/python/src/tiledb/vector_search/object_api/embeddings_ingestion.py
```diff
@@ -2,335 +2,476 @@
...
+            partition_dicts: List[Dict],
+            embeddings_uri: str,
+            metadata_array_uri: str = None,
+            index_timestamp: int = None,
+            verbose: bool = False,
+            trace_id: Optional[str] = None,
+            config: Optional[Mapping[str, Any]] = None,
+            extra_worker_modules: Optional[List[str]] = None,
+        ):
+            def install_extra_modules():
```
jparismorgan, line 137: (Minor)(Optional) Could we rename this to something else so it's easier to understand whether this one or the one above is being called? I.e. `install_extra_worker_modules` (and above could be `install_extra_driver_modules`?

apis/python/test/test_object_index.py
```diff
@@ -145,6 +145,7 @@ def test_object_index_ivf_flat(tmp_path):
     )
 
     index.update_index()
+    index = object_index.ObjectIndex(uri=index_uri)
```
jparismorgan, line 148: (Non-blocking) Could you explain a bit more about why we need to do this? It seems much simpler to not have to?

jparismorgan (approved):

apis/python/src/tiledb/vector_search/object_api/embeddings_ingestion.py
NikolaosPapailiou: Applied

apis/python/test/test_object_index.py
NikolaosPapailiou: We can't pass the initialized `ObjectIndex` to the taskgraph as this would need to pickle all open arrays and memory of the open index class. This is why we just pass the `index_uri` and re-initialize on the Taskgraph side.

PR #258, opened 2024-03-01
Check for a nullptr `index_` in all api/index methods

jparismorgan:
### What
Right now we seg fault if you call APIs in the wrong order. Here we guard against that and instead throw with a more useful error.

### Testing 
* Builds and tests pass.

### DIFF
--- a/src/include/api/flat_l2_index.h
+++ b/src/include/api/flat_l2_index.h
@@ -91,6 +91,9 @@ class IndexFlatL2 {
   // @todo query() or search() -- or both?
   [[nodiscard]] auto query(
       const QueryVectorArray& vectors, size_t top_k) const {
+    if (!index_) {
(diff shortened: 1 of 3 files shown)


ihnorton (approved):
I'm ok with this to fix an immediate bug, however we should revisit ASAP and make sure that objects are usable upon construction (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-complete).

jparismorgan:
> I'm ok with this to fix an immediate bug, however we should revisit ASAP and make sure that objects are usable upon construction (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-complete).

Agree, created a story: https://app.shortcut.com/tiledb-inc/story/42372

PR #259, opened 2024-03-03
Small type alias fix in matrix_with_ids.h

lums658:
This just fixes a few errors / warnings that popped up when compiling with clang-14 on ubuntu.

In general, classes should have a standardized set of internal type aliases (which we haven't really formally done, nor documented the same) -- and those names should be defined at the top of the class and then used in the body of the class.  

In this case, the body of the class was using `Base::size_type` rather than just `size_type` -- the compiler caught it because there wasn't a `typename` there -- but there may be other instances of similar kinds of non-ideal usage.

See C++ Core Guidelines T.42 -- which we should adopt .  The usage that I am suggesting we follow is a slight generalization (but based on same principle).


jparismorgan (approved):

NikolaosPapailiou (approved):

jparismorgan:
Thanks for fixing this @lums658  - I pulled from main and fixed merge conflicts to make this easier to merge.

PR #262, opened 2024-03-05
Add support for TileDB sparse arrays in vector ingestion

NikolaosPapailiou:
This add support for TileDB sparse arrays as input in vector ingestion.

It also:
- Removes the attribute name `values` requirement for dense arrays. Instead we assume that the values are in the first attribute of the array schema.
- Fixes a bug in random sampling where we used `in_size` instead of `size` to compute the sampling tasks.

The sparse array input functionality can be used facilitate the ingestion of single cell census embeddings (https://cellxgene.cziscience.com/census-models)

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -2356,9 +2356,9 @@ def ingest(
         logger.debug("Input dataset size %d", size)
         logger.debug("Input dataset dimensions %d", dimensions)
         logger.debug("Vector dimension type %s", vector_type)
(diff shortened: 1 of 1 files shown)


jparismorgan (approved):
Nice catch on the `size` vs `in_size` fix as well - curious, did we run into an issue because of it? Or you just noticed it?

NikolaosPapailiou:
> Nice catch on the `size` vs `in_size` fix as well - curious, did we run into an issue because of it? Or you just noticed it?

I had some issues with ingesting some test arrays with random sampling

PR #269, opened 2024-03-07
Add support for reading an empty vector with `read_vector()`

jparismorgan:
### What
Right now in `read_vector()` you cannot read an empty vector because we'll automatically read to the domain end if you pass 0. Here we update so we have a different constructor for automatically reading to the end versus reading a range.

### Testing
* Existing tests pass.
* New unit test checking that we can read an empty vector with this function.

### DIFF
--- a/src/include/detail/linalg/tdb_io.h
+++ b/src/include/detail/linalg/tdb_io.h
@@ -43,6 +43,77 @@
 #include "utils/print_types.h"
 #include "utils/timer.h"
 
+namespace {
+
+template <class T>
+std::vector<T> read_vector_helper(
+    const tiledb::Context& ctx,
+    const std::string& uri,
+    size_t start_pos,
+    size_t end_pos,
+    size_t timestamp,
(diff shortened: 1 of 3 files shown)


lums658 (changes requested):
We should use `std::optional` in `read_vector` as well, rather than `0` as a sentinel value.  I think I may have left a comment in that I meant to delete about `0` being okay.  It's important to be consistent between `tdbMatrix` and `read_vector`.

lums658 (approved):
Per discussion between @lums658 and @jparismorgan, the various subtle issues in my comments in related PR #268 regarding start/stop etc are deferred to a subsequent PR.

jparismorgan:
> Per discussion between @lums658 and @jparismorgan, the various subtle issues in my comments regarding start/stop etc are deferred to a subsequent PR.

Thanks, created a story for this here: https://app.shortcut.com/tiledb-inc/story/42770

PR #275, opened 2024-03-11
Remove `[ci-skip]` on unit tests

jparismorgan:
### What
Not sure why we need these, but things pass without it, so remove it.

### Testing
Tests pass.


lums658 (approved):
These were in place because we didn't have test data in a standard place or with any standard naming.  As a result, testing with array data was accomplished with hard-coded paths.  Those tests had to be disabled when running under CI.  Now that we have test data in our repo, we can (and should) use the data in CI.

NikolaosPapailiou (approved):

PR #276, opened 2024-03-12
Fix packaging version

dudoslav:
This PR fixes version not correctly exporting in wheels. This is the current output if build locally:
```
Successfully installed tiledb-vector-search-0.1.1.dev25+gb0c5208e.d20240312
(zoo) taco:TileDB-Vector-Search dudoslav$ python
Python 3.11.7 | packaged by conda-forge | (main, Dec 23 2023, 14:43:09) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from tiledb.vector_search.version import version
>>> version
'0.1.1.dev25+gb0c5208e.d20240312'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,12 +43,12 @@ cmake.source-dir = "src"
 cmake.targets = ["skbuild-export"]
 # TODO: This should be removed, but currently can't bacause: 'src/src/CMakeLists.txt'
 sdist.include = [
-    ".git"
+    ".git",
+    "apis/python/src/tiledb/vector_search/version.py"
 ]
 wheel.packages = ["apis/python/src/tiledb"]
 
 [tool.scikit-build.cmake.define]
-#GIT_REPO_NAME = "placeholder"
 TILEDB_VS_PYTHON = "ON"
 
 [tool.setuptools_scm]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,12 +41,17 @@ wheel.expand-macos-universal-tags = true
 metadata.version.provider = "scikit_build_core.metadata.setuptools_scm"
 cmake.source-dir = "src"
 cmake.targets = ["skbuild-export"]
-# TODO: This should be removed, but currently can't bacause: 'src/src/CMakeLists.txt'
 sdist.include = [
-    ".git",
-    "apis/python/src/tiledb/vector_search/version.py"
+    "apis/python/src/tiledb/vector_search/version.py",
+    "src/config.h"
+]
+sdist.exclude = [
+    ".github"
 ]
 wheel.packages = ["apis/python/src/tiledb"]
+# We need to run cmake during packaging step in order to generate src/config.h and
+# thus not including .git to generate it later
+sdist.cmake = true
 
 [tool.scikit-build.cmake.define]
 TILEDB_VS_PYTHON = "ON"
--- a/src/cmake/logging.cmake
+++ b/src/cmake/logging.cmake
@@ -1,5 +1,4 @@
-
-if(NOT EXISTS ${CMAKE_SOURCE_DIR}/config.h)
+if(EXISTS ${CMAKE_SOURCE_DIR}/../.git)
 
 # Get the current date and time
 string(TIMESTAMP CURRENT_DATETIME "%Y-%m-%d %H:%M:%S")
@@ -60,4 +59,8 @@ execute_process(
 
 get_filename_component(IVF_HACK_CXX_COMPILER ${CMAKE_CXX_COMPILER} NAME)
 
+set(LOGGING_INFO_QUERIED)
+
+elseif(NOT EXISTS ${CMAKE_SOURCE_DIR}/config.h)
+    message(FATAL_ERROR ".git or pre generated config.h is required")
 endif()
--- a/src/cmake/logging.cmake
+++ b/src/cmake/logging.cmake
@@ -1,4 +1,5 @@
 
+if(NOT EXISTS ${CMAKE_SOURCE_DIR}/config.h)
 
 # Get the current date and time
 string(TIMESTAMP CURRENT_DATETIME "%Y-%m-%d %H:%M:%S")
@@ -58,3 +59,5 @@ execute_process(
 )
 
 get_filename_component(IVF_HACK_CXX_COMPILER ${CMAKE_CXX_COMPILER} NAME)
+
+endif()
--- a/src/cmake/logging.cmake
+++ b/src/cmake/logging.cmake
@@ -59,7 +59,7 @@ execute_process(
 
 get_filename_component(IVF_HACK_CXX_COMPILER ${CMAKE_CXX_COMPILER} NAME)
 
-set(LOGGING_INFO_QUERIED)
+set(LOGGING_INFO_QUERIED True)
 
 elseif(NOT EXISTS ${CMAKE_SOURCE_DIR}/config.h)
     message(FATAL_ERROR ".git or pre generated config.h is required")
--- a/.github/workflows/ci-python.yml
+++ b/.github/workflows/ci-python.yml
@@ -40,3 +40,6 @@ jobs:
         env:
           TILEDB_REST_TOKEN: ${{ secrets.TILEDB_CLOUD_HELPER_VAR }}
         shell: bash -el {0}
+      - name: Check tiledb-vector-search version
+        run: |
+          python -c "from tiledb.vector_search.version import version; print(version)"
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -175,7 +175,7 @@ include(nlohmann_json)
 include (logging)
 
 # Write config file
-if(NOT EXISTS ${CMAKE_SOURCE_DIR}/config.h)
+if(LOGGING_INFO_QUERIED)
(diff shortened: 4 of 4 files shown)


pyproject.toml
```diff
@@ -43,12 +43,12 @@ cmake.source-dir = "src"
 cmake.targets = ["skbuild-export"]
 # TODO: This should be removed, but currently can't bacause: 'src/src/CMakeLists.txt'
 sdist.include = [
-    ".git"
+    ".git",
```
ihnorton, line 46: Hmm, I would rather not include this in the sdist. Please try this instead: https://github.com/TileDB-Inc/TileDB-Py/blob/ff2149738a5daeee2c11d7660b3c7631fe1a293e/setup.py#L809

dudoslav: I would also like to remove it from there, but the reason why I left .git in this include is that the CMake that we use actually use it. Not setuptools_scm. And that logic in cmake breaks during sdist build, since it looks for .git folder. However, I just noticed that that git logic got moved to `logging.cmake` and it might not even be used anymore. I'll try `sdist` and if it works I'll remove that .git folder from package.

dudoslav:
```
      FAILED: libtiledbvectorsearch-prefix/src/libtiledbvectorsearch-stamp/libtiledbvectorsearch-configure /tmp/tmpji08b7ja/build/libtiledbvectorsearch-prefix/src/libtiledbvectorsearch-stamp/libtiledbvectorsearch-configure
      cd /tmp/tmpji08b7ja/build/libtiledbvectorsearch && /usr/bin/cmake -DSUPERBUILD=OFF -DCMAKE_INSTALL_PREFIX=/usr/local "-DCMAKE_PREFIX_PATH=/home/dudoslav/.miniforge3/envs/zoo/lib/python3.11/site-packages;/tmp/pip-build-env-0umpripn/overlay/lib/python3.11/site-packages;" -DCMAKE_MODULE_PATH=/tmp/pip-req-build-9ud69_n7/src/cmake/Modules/ /tmp/pip-req-build-9ud69_n7/src/cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_SYSTEM_PROCESSOR=x86_64 -DCMAKE_OSX_ARCHITECTURES= -DCMAKE_C_COMPILER=/usr/bin/cc -DCMAKE_C_FLAGS= -DCMAKE_CXX_COMPILER=/usr/bin/c++ -DCMAKE_CXX_FLAGS= -DCMAKE_CXX_STANDARD=20 -DEP_BASE=/tmp/tmpji08b7ja/build/externals -DFORCE_EXTERNAL_TILEDB=OFF -DTILEDB_S3=ON -DTileDB_DIR= -DOVERRIDE_INSTALL_PREFIX=OFF -DBLA_VENDOR= -DUSE_MKL_CBLAS=OFF -DTILEDB_VS_PYTHON=ON -DBUILD_TESTS=ON -DPython_FIND_REGISTRY=NEVER -DPYTHON_VERSION_STRING= -DPYTHON_LIBRARY= -DPYTHON_INCLUDE_DIR=/home/dudoslav/.miniforge3/envs/zoo/include/python3.11 -DPython_EXECUTABLE=/home/dudoslav/.miniforge3/envs/zoo/bin/python3.11 -DPython_ROOT_DIR=/home/dudoslav/.miniforge3/envs/zoo -DPython_INCLUDE_DIR=/home/dudoslav/.miniforge3/envs/zoo/include/python3.11 -DPython3_FIND_REGISTRY=NEVER -DPython3_INCLUDE_DIR=/home/dudoslav/.miniforge3/envs/zoo/include/python3.11 -Dpybind11_DIR= -DEP_TILEDB_BUILT=TRUE -GNinja -S /tmp/pip-req-build-9ud69_n7/src -B /tmp/tmpji08b7ja/build/libtiledbvectorsearch && /usr/bin/cmake -E touch /tmp/tmpji08b7ja/build/libtiledbvectorsearch-prefix/src/libtiledbvectorsearch-stamp/libtiledbvectorsearch-configure
      CMake Warning:
        Ignoring extra path from command line:
      
         "/tmp/pip-req-build-9ud69_n7/src/cmake"
      
      
      CMake Warning:
        Ignoring extra path from command line:
      
         "/tmp/pip-req-build-9ud69_n7/src/cmake"
      
      
      -- The C compiler identification is GNU 13.2.1
      -- The CXX compiler identification is GNU 13.2.1
      -- Detecting C compiler ABI info
      -- Detecting C compiler ABI info - done
      -- Check for working C compiler: /usr/bin/cc - skipped
      -- Detecting C compile features
      -- Detecting C compile features - done
      -- Detecting CXX compiler ABI info
      -- Detecting CXX compiler ABI info - done
      -- Check for working CXX compiler: /usr/bin/c++ - skipped
      -- Detecting CXX compile features
      -- Detecting CXX compile features - done
      -- Starting TileDB-Vectorsearch regular build.
      Architecture is x86_64
      -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
      -- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
      -- Found Threads: TRUE
      -- TileDB_DIR is set to  -- find_package will search there first.
      -- Found TileDB: /tmp/tmpji08b7ja/build/externals/install/lib/libtiledb.so.2.20
      CMake Deprecation Warning at /tmp/tmpji08b7ja/build/libtiledbvectorsearch/_deps/docopt-src/CMakeLists.txt:1 (cmake_minimum_required):
        Compatibility with CMake < 3.5 will be removed from a future version of
        CMake.
      
        Update the VERSION argument <min> value or use a ...<max> suffix to tell
        CMake that the project does not need compatibility with older versions.
      
      
      -- Fetching mdspan...
      -- Detected support for C++23 standard
      -- Performing Test COMPILER_SUPPORTS_FCONCEPTS
      -- Performing Test COMPILER_SUPPORTS_FCONCEPTS - Success
      -- Using "-fconcepts" to enable concepts support
      Installed json not found -- fetching
      CMake Deprecation Warning at /tmp/tmpji08b7ja/build/libtiledbvectorsearch/_deps/json-src/CMakeLists.txt:1 (cmake_minimum_required):
        Compatibility with CMake < 3.5 will be removed from a future version of
        CMake.
      
        Update the VERSION argument <min> value or use a ...<max> suffix to tell
        CMake that the project does not need compatibility with older versions.
      
      
      -- Using the single-header code from /tmp/tmpji08b7ja/build/libtiledbvectorsearch/_deps/json-src/single_include/
      fatal: not a git repository (or any parent up to mount point /)
      Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
      fatal: not a git repository (or any parent up to mount point /)
      Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
      CMake Error at cmake/logging.cmake:28 (get_filename_component):
        get_filename_component called with incorrect number of arguments
      Call Stack (most recent call first):
        CMakeLists.txt:175 (include)
      
      
      -- Configuring incomplete, errors occurred!
      ninja: build stopped: subcommand failed.
      
      *** CMake build failed
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for tiledb-vector-search
Failed to build tiledb-vector-search
ERROR: Could not build wheels for tiledb-vector-search, which is required to install pyproject.toml-based projects
```

dudoslav: This is what happens ^^ when I remove `.git` from that repo. It will break `sdist`

ihnorton: We know we can generate a version file from setuptools, so I think we should do that and then parse it in cmake to get the version (if this file exists).

Some related discussion here, but the suggestion is a separate repo AFAICT: https://github.com/scikit-build/scikit-build-core/issues/631

dudoslav: This is a great observation, and yes, we could use `SKBUILD_PROJECT_VERSION_FULL`. But unfortunately, right now we log these values:
```
constexpr const char GIT_REPO_NAME[] = "@GIT_REPO_NAME@";
constexpr const char GIT_REPO_URL[] = "@GIT_REPO_URL@";
constexpr const char GIT_BRANCH[] = "@GIT_BRANCH@";
constexpr const char GIT_COMMIT_HASH[] = "@GIT_COMMIT_HASH@";
constexpr const char GIT_COMMIT_DATE[] = "@GIT_COMMIT_DATE@";
constexpr const char GIT_COMMIT_TIME[] = "@GIT_COMMIT_TIME@";
```

That are not present in that previously mentioned variable.

dudoslav: I would suggest merging this PR in order to fix the versioning ASAP and create a new story just for removing `.git` from sdist.

src/cmake/logging.cmake
```diff
@@ -1,4 +1,5 @@
 
+if(NOT EXISTS ${CMAKE_SOURCE_DIR}/config.h)
```
ihnorton, line 2: Should this be unconditional in case there are changes since the last run?

dudoslav: Ah, well, very good point. The logic here has to prevent any generation/git querying if we are building from sdist, so we don't have `.git` in folder. I have another idea, I can check if there is .git folder and only generate that file in that case.

ihnorton: SGTM thanks

ihnorton (approved):

PR #278, opened 2024-03-12
Add `has_updates` to C++ metadata

jparismorgan:
### What
Here we add `has_updates` to C++ metadata - without it we crash when querying with the type-erased Vamana index. The one interesting thing to me is that in Python `has_updates` is an int64:
```
### Array Metadata ###
- Key: has_updates
- Value: 0
- Type: DataType.INT64
```
So we also do that in C++.

### Changes
In order to get this to work I made two changes beyond the core code and unit tests:
* Update the Vamana index in `external/test_data/nano/vamana/vamana_test_index`, as well as `src/include/test/notebooks/gen_vamana_metadata.ipynb` which is used to generate it. 
  * Side note: I don't like this pattern of checking in extra arrays to the repo because all they really test for is backwards compatibility. It seems better to instead use the backwards compatibility CI job and indexes it creates, instead of these one-off indexes. This is easier to maintain and simpler, yet still provides similar testing coverage for the same set of potential issues. In a follow-up I plan to remove these indexes, but I'll hold off for now.
* Add `#include "detail/linalg/vector.h"` to `src/include/index/vamana_index.h` - we failed to compile in one of the unit tests without it.

I also added some unit tests to `apis/python/test/test_index.py` that check the metadata when the indexes are created. This is not required, but I used it to understand the Python code better, and it seems like a nice test to add.

### TODO
* As mentioned above, in a future update I would like to remove as man indexes checked into `external/test_data/` as I can.
* Once I do that I will be able to mark `has_updates` as a required field in `src/include/index/vamana_index.h`. Right now it is optional because otherwise all the unit tests would fail. I could update the checked in indexes (I did ask Andrew to share the code he used to generate them), but I still don't love that, as it just means the next time we want to change something we need to regenerate the indexes.

### Testing
* Adds several new unit tests. One of these is to explicitly load metadata in `src/include/test/unit_backwards_compatibility.cc`. This check was then removed from `src/include/test/unit_ivf_flat_metadata.cc` and `src/include/test/unit_vamana_metadata.cc`. In it's place I added a check where we create an index and then load the metadata.
* Olds unit tests pass.


jparismorgan:
Talked to @NikolaosPapailiou offline - we will want this eventually, but for now we can just make Python work with C++ indexes which do not have `has_updates`. Then we need to change less code which is nice. So closing for now.

PR #282, opened 2024-03-14
Optimized graph algorithms for vamana

lums658:
This draft PR is an experimental branch that includes a number of new algorithms for performing the truncated best-first search used by vamana for queries and for graph construction.  The main purpose of experimenting with the different algorithms was to bring our performance on-par with Microsoft's DiskANN implementation.  The search algorithms are included in files in detail/graph: bfs.h. dfs.h, greedy-search.h, and best_first.h.

The experiments included bfs and dfs in order to understand simple baseline performance for traversing the graph structure we have.  The original search algorithm, `greedy_search` was implemented based on the DiskANN paper.  The algorithms in best_first.h were implemented from scratch.  The algorithms `best_first_O0` through `best_first_O5` use different approaches for maintaining the frontier of neighbors to be explored as the search progresses, as well as different structures to represent the state of nodes during the search.  `best_first_O4` seems to offer the best performance and is what the query function in `vamana_index` dispatches to (or is what it should dispatch to).

The basic idea of the best_first algorithms in this PR is to maintain a set of vertices as the frontier to be explored.  As with typical graph traversal algorithms, at each iteration we select a vertex from the current frontier and selectively add its neighbors to the current frontier in order to create the next frontier.  In the case of best-first, we choose the vertex closest to the query as the one next to explore.  At the end of the iteration, we also trim the frontier to be the L closest vertices to the query vector.

One of the key observations leading to the improved performance of O4 is that the frontier is implicit -- we don't need to maintain one explicitly in order to be able to "extract" the vertex nearest to the query.  This should be applicable to robust prune as well.

The CLI programs in src/src/vamana support the new algorithms -- vamana_query.cc takes options that allow the user to specify which query algorithm to use -- and it prints out various performance data.  The scripts in src/benchmarks/ can be executed to run the Microsoft programs (diskann.bash) or to run our programs (vamana.bash).  These have all the flags necessary to set up the run and to output the timing results.

vamana uses best-first search in its queries -- it also uses it as part of index construction, along with "robust prune".  Although best-first search has been optimized to match the performance of DiskANN, robust prune has not.  This will be important to do to get indexing performance comparable to DiskANN.  The robust prune algorithm is very similar to a traversal so we should readily be able to write a version with performance that matches DiskANN, using the techniques we used for best-first search.  (It will already have a big boost due to improved distance function performance due to avx2 instructions.)

The vamana indexing algorithm we implement is taken from the most recent DiskANN paper (filtered DiskANN), which is a much more efficient approach than in the first few papers.  Our current implementation is overly conservative (to help with debugging and error checking).  This can likewise be significantly improved.

**NOTE** the query function for vamana is currently sequential and our performance comparisons with DiskANN were for single-threaded performance (which is also what the bigann suite will use).  But to match DiskANN in general, we will need to multithread the vamana query function (there is a story for that and it should be straightforward to do).

We will also want to parallelize the indexing process -- the current algorithm is sequential.  This will require some care given the unstructured nature of the graph as it is being constructed.  We want to maximize concurrency without too much fine-grained locking.

### Testing
A fair amount of testing was done manually, based on running the various versions of vamana from the command line and comparing performance and accuracy vs DiskANN.  Our code can also read in a graph created by DiskANN so we checked our accuracy using `sift` and `siftsmall` graphs generated by DiskANN.  Our programs gave exactly the same accuracy.  The indexes we built with our own implementation gave a slightly better accuracy than DiskANN.


src/include/detail/graph/adj_list.h
```diff
@@ -143,7 +143,7 @@ class adj_list : public std::vector<std::list<std::tuple<SC, ID>>> {
   adj_list(AdjList&& l)
       : Base(l.size()) {
     for (size_t i = 0; i < l.size(); ++i) {
-      for (auto&& [dst, val] : l[i]) {
+      for (auto&& [val, dst] : l[i]) {
         add_edge(i, dst, val);
       }
     }
```
jparismorgan, line 149: @lums658 how was our code working before this? Should unit tests have been failing? If not, are there unit tests we can write that will fail if we change this back to the incorrect version?

src/include/test/CMakeLists.txt
```diff
@@ -153,6 +153,7 @@ kmeans_add_test(unit_tdb_matrix)
 kmeans_add_test(unit_tdb_matrix_with_ids)
 
 kmeans_add_test(unit_utils)
+set_property(TEST unit_utils_test PROPERTY WILL_FAIL TRUE)
```
jparismorgan, line 156: I do not want to move forward like this without more understanding of why, so let's not merge this PR until we've discussed more.

src/include/detail/graph/adj_list.h
lums658: Unit test added.

src/include/test/CMakeLists.txt
lums658: Updated unit test to use different known file.  Removed `WILL_FAIL`.

src/include/detail/graph/dfs.h
```diff
@@ -0,0 +1,36 @@
+/**
+ * @file   dfs.h
```
jparismorgan, line 2: (Minor) Let's not check this in if it's just an empty file.

src/include/detail/graph/diskann.h
```diff
@@ -129,21 +129,31 @@ auto read_diskann_mem_index_with_scores(
...
     for (size_t i = 0; i < num_neighbors; ++i) {
       uint32_t id;
       binary_file.read((char*)&id, 4);
+      assert(id < num_nodes);
       g.add_edge(node, id, l2_distance(x[node], x[id]));
     }
     // @todo ??? Is this right ???
-    binary_file.seekg(max_degree - num_neighbors, std::ios_base::cur);
+    // binary_file.seekg(max_degree - num_neighbors, std::ios_base::cur);
+    ++node;
```
jparismorgan, line 153: @lums658 Can you explain more about what's going on here? Can we remove these lines?
```
    // @todo ??? Is this right ???
    // binary_file.seekg(max_degree - num_neighbors, std::ios_base::cur);
```

src/include/detail/graph/greedy_search.h
```diff
@@ -0,0 +1,550 @@
...
+      std::vector<typename std::decay_t<decltype(graph)>::id_type>(k_nn);
+  auto top_k_scores = std::vector<float>(k_nn);
+
+  get_top_k_with_scores_from_heap(min_scores, top_k, top_k_scores);
+  return std::make_tuple(
+      std::move(top_k_scores), std::move(top_k), std::move(V));
+
+  return std::make_tuple(
+      std::move(top_k_scores), std::move(top_k), std::move(V));
+}
```
jparismorgan, line 237: (Minor) Duplicate return

src/include/detail/graph/greedy_search.h
```diff
@@ -0,0 +1,550 @@
...
+#include "scoring.h"
+#include "utils/fixed_min_heap.h"
+
+static bool noisy = false;
+[[maybe_unused]] static void set_noisy(bool b) {
+  noisy = b;
+}
+
+template </* SearchPath SP, */ class Distance = sum_of_squares_distance>
+auto greedy_search_O2(
```
jparismorgan, line 52: Would be good to give these descriptive names which explain the differences.

src/include/detail/graph/greedy_search.h
```diff
@@ -0,0 +1,550 @@
...
+template <class I = size_t, class Distance = sum_of_squares_distance>
+auto robust_prune(
+    auto&& graph,
+    auto&& db,
+    I p,
+    auto&& V_in,
+    float alpha,
+    size_t R,
+    Distance&& distance = Distance{}) {
+  constexpr bool noisy = false;
```
jparismorgan, line 458: (Minor) Should this use the top-level `noisy`?

src/include/test/unit_vamana_index.cc
```diff
@@ -325,7 +334,7 @@ TEST_CASE("vamana: small greedy search", "[vamana]") {
   // L = 50, R = 4
   size_t L = 45;
   auto query_id = 14;
-  size_t k = 15;
+  size_t k = 2;
```
jparismorgan, line 337: @lums658 Does this fail with `k = 15`? Curious, why change it?

src/include/test/unit_vamana_index.cc
```diff
@@ -656,13 +703,16 @@ TEST_CASE("vamana: diskann fbin", "[vamana]") {
     // the start point should be included in the top_k list.  It probably
     // should not be, so comment out for now.
     // CHECK(top_k[0] == start);
-    CHECK(std::find(begin(top_k), end(top_k), start) != end(top_k));
-    CHECK(std::find(begin(V), end(V), start) != end(V));
+    // CHECK(std::find(begin(top_k), end(top_k), start) != end(top_k));
+    // CHECK(std::find(begin(V), end(V), start) != end(V));
```
jparismorgan, line 707: @lums658 Can we add this back? Why remove these checks?

src/include/utils/minmax_and_dary_heap.h
```diff
@@ -0,0 +1,722 @@
...
+ *
+ * Implementation of a min-max heap, taken from https://github.com/skarupke/heap
+ * Copyright Malte Skarupke 2020.
+ * Distributed under the Boost Software License, Version 1.0.
+ * (See http://www.boost.org/LICENSE_1_0.txt)
+ *
+ */
+
+#ifndef TILEDB_MIN_MAX_HEAP_H
+#define TILEDB_MIN_MAX_HEAP_H
```
jparismorgan, line 38: I don't see anything referencing this file - am I just missing the use of it? Or do we not need to add it?

Would also be good to add a few tests.

src/include/detail/graph/greedy_search.h
lums658: Created a separate `noisy_robust_prune` to separate noisy `greedy_search` from `robust_prune`

src/include/detail/graph/greedy_search.h
lums658: O0, O1, O2 were different approaches for accelerating greedy_search.  `best_first` has superseded `greedy_search` so we may just want to deprecate / remove `greedy_search` altogether.

src/include/utils/minmax_and_dary_heap.h
lums658: I was experimenting with it in `greedy_search` and `best_first`.  The annoying thing is that if you keep your frontier in a priority queue, you need to access both the min element and the max element at different points in the algorithm. However, I realized at some point that the queue doesn't have to be materialized and so we don't actually need the `min_max_heap`.  Deleted.

src/include/test/unit_vamana_index.cc
lums658: I actually don't know why it was changed.  I changed it back to 15.  There was a typo in one of the expected values, which I fixed.  The test passes.

src/include/test/unit_vamana_index.cc
lums658: They were commented out pending determination of whether or not the start should be included (to be consistent with DiskANN).  Uncommented.

src/include/detail/graph/dfs.h
jparismorgan: @lums658 it looks like this was not deleted, though this comment was marked resolved. Could we delete? 

Also, in the future, would you mind leaving all my comments unresolved? I will go through them myself and mark as resolved. This helps avoid situations like this where things are incorrectly marked as resolved.

src/include/test/unit_vamana_index.cc
jparismorgan: Great, thanks.

src/include/test/unit_vamana_index.cc
jparismorgan: Great, thanks

src/include/detail/graph/dfs.h
lums658: Roger.

src/include/utils/minmax_and_dary_heap.h
lums658: Yes.

jparismorgan:
NOTE(paris): I'm currently waiting to merge this until Vamana is fully working end to end, as that will decrease the risk that a bug is introduced in this PR and not caught with testing.

jparismorgan (approved):

PR #283, opened 2024-03-14
Add support for more input sources in DirectoryReader

NikolaosPapailiou:
This adds support for listing TileDB groups

It also uses the same listing argument structure as the utilities used in VCF ingestion

[sc-42043]


apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +8,223 @@
...
+                    search_uri="/",
+                    file_uri=path,
+                    include=include,
+                    exclude=exclude,
+                    suffixes=suffixes,
+                )
+                if match:
+                    results.append(uri)
+            elif obj_type == "group":
+                grp = tiledb.Group(uri)
```
Shelnutt2, line 133: Perhaps use a context manager (`with`), or explicitly close the group after listing?

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +8,223 @@
...
+            suffixes=suffixes,
+            max_files=max_files,
+        )
+        if max_files is not None:
+            return results[:max_files]
+        else:
+            return results
+
+
+def find_uris_aws(
```
Shelnutt2, line 159: We should avoid this and use VFS and vfs recursive listing so we are cross platform compatible.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -67,25 +287,34 @@ def partition_class_name(self) -> str:
...
-                if not child_path.match(self.glob):
-                    continue
-
-                if self.suffixes and child_path.suffix not in self.suffixes:
-                    continue
-                self.paths.append(child)
-            else:
-                self.list_paths(vfs, child)
+    def list_paths(self):
+        if self.search_uri.startswith("s3://"):
```
Shelnutt2, line 291: We should use VFS for everything and not use boto3.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +8,223 @@
...
+            exclude: Sequence[str] = (),
+            suffixes: Optional[Sequence[str]] = None,
+            max_files: Optional[int] = None,
+        ):
+            # Stop if we have found max_count files
+            if max_files is not None and len(results) >= max_files:
+                return
+            obj_type = tiledb.object_type(uri)
+            if obj_type == "array":
+                match = match_uri(
```
Shelnutt2, line 123: Question to make sure I understand the logic, is this matching requiring the array names to match expected file name?

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -178,7 +409,18 @@ def read_objects(
                 try:
                     local_temp_path = ""
                     parsed_path = urlparse(path, allow_fragments=False)
-                    if parsed_path.scheme != "file":
+                    if parsed_path.scheme == "tiledb":
+                        key = parsed_path.path.lstrip("/")
+                        with tiledb.open(path, "r") as a:
+                            file_extension = a.meta["file_extension"].lstrip(".")
```
Shelnutt2, line 415: I believe you need some checks and validation that this is a TileDB File or these accesses with throw.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -178,7 +409,18 @@ def read_objects(
...
+                            file_size = a.meta["file_size"]
+                            local_temp_path = f"{temp_dir}/{key}.{file_extension}"
+                            os.makedirs(os.path.dirname(local_temp_path), exist_ok=True)
+                            data = a[:file_size]["contents"]
+                            with open(local_temp_path, "wb") as dst:
```
Shelnutt2, line 420: Can we avoid copying the file locally? Previously I was able to write a simple class for langchain to extend byte reading to TileDB datasets directly with no copy, we can probably maintain simple support here? We strive in general for zero copy and direct in memory integration so I'd like to see what we can do here.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -178,7 +409,18 @@ def read_objects(
...
+                                dst.write(data)
+                        parsed_path = urlparse(local_temp_path, allow_fragments=False)
+                    if parsed_path.scheme == "s3":
```
Shelnutt2, line 423: Can we do the same with s3 to avoid copying? Copying anything should be avoided and removed if possible.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Applied

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: This is added for performance reasons. Listing large S3 directories and checking individual files using VFS takes a lot of time. This is also similar to the approach used in [VCF](https://github.com/TileDB-Inc/TileDB-Cloud-Py/blob/main/src/tiledb/cloud/vcf/ingestion.py#L438) for listing S3. I avoided using the `aws` command line util here and switched to `boto3`.

We could add an option to use `find_uris_vfs` for S3 if you think there is going to be a compatibility issue.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Commented on this above.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: No this is matching the full path of the file inside the group structure.

We maintain the path as we follow the different groups in order to support similar matching functionality as the other directory based functions.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Langchain readers require the file to be in the local filestystem. I haven't found a way to pass the file handler directly. 

Maybe a solution would be to write the files in the UDF memory based filesystem. We can also add a separate task to explore this more.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: See comment above

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Do we have a consistent way to check for file type assets? I couldn't find a type metadata tag. Should we just check for the existence of `file_extension, file_size`?

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
Shelnutt2: If there is performance, open a item and lets fix things. You can also use the ls recursive support. VCF does not use aws cli or boto3 by default and was migrated away from it with plans to remove it. Its unused code by default.

As the `find_uris_aws` is written today it also has issues with chaining roles and using a aws file config. The core library supports all of this, lets avoid re-implementing specific functionality already covered in a cross platform way by the core library.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
Shelnutt2:
```
class TileDBPyPDFLoader(BasePDFLoader):
        """Load PDF using pypdf into list of documents.

        Loader chunks by page and stores page numbers in metadata.
        """

        def __init__(
            self,
            file_path: str,
            password: Optional[Union[str, bytes]] = None,
            headers: Optional[Dict] = None,
            extract_images: bool = False,
            ctx: Optional[tiledb.Ctx] = None,
            config: Optional[Dict] = None,
        ) -> None:
            """Initialize with a file path."""
            try:
                import pypdf  # noqa:F401
            except ImportError:
                raise ImportError(
                    "pypdf package not found, please install it with " "`pip install pypdf`"
                )
            self.file_path = file_path
            self.ctx = ctx
            self.config = config
            self.parser = PyPDFParser(password=password, extract_images=extract_images)

        def load(self) -> List[Document]:
            """Load given path as pages."""
            return list(self.lazy_load())

        def lazy_load(
            self,
        ) -> Iterator[Document]:
            """Lazy load given path as pages."""
            with tiledb.scope_ctx(self.ctx):
                if tiledb.array_exists(self.file_path):
                    f = tiledb.filestore.Filestore(self.file_path)
                    blob = Blob.from_data(f.read())
                else:
                    vfs = tiledb.VFS(config=self.config, ctx=self.ctx)
                    with vfs.open(self.file_path) as f:
                        blob = Blob.from_data(f.read())
            yield from self.parser.parse(blob)
```

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Used `find_uris_vfs` for S3. 

What is the vfs ls recursive support? I only see non-recursive ls if TileDB-Py https://github.com/TileDB-Inc/TileDB-Py/blob/5cfefdd0d72a90fa4e16fb2f1107cb09a86b2958/tiledb/vfs.py#L290

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Created https://app.shortcut.com/tiledb-inc/story/43213/avoid-copying-files-locally-in-udfs-during-vector-store-file-indexing to track this separately as this will need some effort to implement and combine with the mimetype-based parsers we are currently using.

NikolaosPapailiou: Applied

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Removed

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +18,223 @@
...
+
+def find_uris_vfs(
+    search_uri: str,
+    config: Optional[Mapping[str, Any]] = None,
+    include: str = "*",
+    exclude: Sequence[str] = ["[.]*", "*/[.]*"],
+    suffixes: Optional[Sequence[str]] = None,
+    max_files: Optional[int] = None,
+) -> Sequence[str]:
+    with tiledb.scope_ctx(config):
```
Shelnutt2, line 53: I believe two contexts are being created here and two lines below. Perhaps it'd be better to create one ctx at the top of the stack and pass that to all functions. Generally we want to avoid repeated context creation as each context contains its own threadpools and other resources.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +18,223 @@
...
+            path: str,
+            include: str = "*",
+            exclude: Sequence[str] = (),
+            suffixes: Optional[Sequence[str]] = None,
+            max_files: Optional[int] = None,
+        ):
+            # Stop if we have found max_count files
+            if max_files is not None and len(results) >= max_files:
+                return
+            obj_type = tiledb.object_type(uri)
```
Shelnutt2, line 131: This needs a scoped context. Likely taking a context as a function parameter would work to pass it as a scoped ctx.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +18,223 @@
...
+            if obj_type == "array":
+                match = match_uri(
```
Shelnutt2, line 133: Matching on name probably won't yield the results expected. We might need to expand this in the future for other filtering like array metadata or original file name/mime type. For now we can leave as is.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +18,223 @@
...
+                    search_uri="/",
+                    file_uri=path,
+                    include=include,
+                    exclude=exclude,
+                    suffixes=suffixes,
+                )
+                if match:
+                    results.append(uri)
+            elif obj_type == "group":
+                with tiledb.Group(uri) as grp:
```
Shelnutt2, line 143: Context needs to be passed.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -35,30 +262,33 @@ def id(self) -> int:
...
-        uri: str,
-        glob: str = "**/[!.]*",
-        exclude: Sequence[str] = (),
-        suffixes: Optional[Sequence[str]] = None,
+        search_uri: str,
         config: Optional[Mapping[str, Any]] = None,
+        include: str = "*",
+        exclude: Sequence[str] = ["[.]*", "*/[.]*"],
+        suffixes: Optional[Sequence[str]] = None,
+        max_files: Optional[int] = None,
```
Shelnutt2, line 270: I believe we need to accept a context or config here in the class constructor for use in subsequent TileDB calls.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -67,25 +297,25 @@ def partition_class_name(self) -> str:
...
-                if not child_path.match(self.glob):
-                    continue
-
-                if self.suffixes and child_path.suffix not in self.suffixes:
-                    continue
-                self.paths.append(child)
-            else:
-                self.list_paths(vfs, child)
+    def list_paths(self):
+        if self.search_uri.startswith("tiledb://"):
```
Shelnutt2, line 301: Should this be a tiledb.object_type check?

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -115,39 +344,107 @@ def get_partitions(
         return partitions
 
 
+class TileDBLoader:
+    def __init__(
+        self,
+        uri: str,
+        config: Optional[Mapping[str, Any]] = None,
```
Shelnutt2, line 351: I believe we should also support passing a context so it can be reused.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -115,39 +344,107 @@ def get_partitions(
...
+
+    def lazy_load(
+        self,
+    ) -> Iterator:
+        """A lazy loader for Documents."""
+        import mimetypes
+
+        from langchain_community.document_loaders.blob_loaders import Blob
+
+        vfs = tiledb.VFS(config=self.config)
```
Shelnutt2, line 388: Context/scope context are needed for all the TileDB calls here and below.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -115,39 +344,107 @@ def get_partitions(
...
+                mime_type = a.meta.get("mime_type", None)
+            f = tiledb.filestore.Filestore(self.uri)
+        else:
+            mime_type = mimetypes.guess_type(self.uri)[0]
+            f = vfs.open(self.uri)
+
+        if mime_type.startswith("image/"):
+            from langchain_community.document_loaders import UnstructuredFileIOLoader
+
+            unstructured_loader = UnstructuredFileIOLoader(f)
```
Shelnutt2, line 400: Does the file store implement the FileIO functions? Can this just be passed to the UnstructuredFileLoader?

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -161,101 +458,35 @@ def metadata_attributes(self) -> List[tiledb.Attr]:
...
-                    import importlib
-
-                    text_splitters_module = importlib.import_module(
-                        "langchain.text_splitter"
+        text_splitters_module = importlib.import_module("langchain.text_splitter")
+        text_splitter_class_ = getattr(text_splitters_module, self.text_splitter)
+        text_splitter = text_splitter_class_(**self.text_splitter_kwargs)
+        for uri in partition.paths:
+            try:
+                loader = TileDBLoader(uri=uri, config=self.config)
```
Shelnutt2, line 475: Shall we pass a context here for reuse?

Shelnutt2:
Great works, thanks for the changes! I've left some minor comments, mostly around the context handling.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +18,223 @@
 # from tiledb.vector_search.object_readers import ObjectPartition, ObjectReader
 
 
+# Util functions for file matching and recursive listing
+def match_uri(
```
jparismorgan, line 22: (Minor)(Optional) Could add unit tests for `match_uri()` / `find_uris_vfs()` / `find_uris_tiledb_group()` etc.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
```diff
@@ -8,6 +18,223 @@
...
+) -> Sequence[str]:
+    from urllib.parse import urlparse
+
+    try:
+        import boto3
+    except ImportError:
+        raise ImportError(
+            "Could not import boto3 python package. "
+            "Please install it with `pip install boto3`."
+        )
```
jparismorgan, line 186: (Curious) This is nice, but, curious, why do it only for `boto3`? Is it because it's a more uncommon package? Or something else I'm missing here?

jparismorgan (approved):
Looks good to me, though you may also want to confirm with Seth.

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Removed all config and context passing from `DirectoryReaders`. 

Readers shouldn't own or manage the TileDB configuration. Rather the config is passed through the ingestion or index functions. Some reason for this:
- Reader arguments are serialized and stored in group metadata. Serializing TileDB config would present a risk of exposing user secrets.
- Multiple users can use the same index with different configs (i.e. each user having their credentials for s3). The Reader and its arguments are the same for all users of an Index

How it works currently:
- Config is passed to `ObjectIndex` methods (`__init__`, `create`, `update_index`).
- We perform all reader operations within `with tiledb.scope_ctx(ctx_or_config=config)`
- Readers should expect a `default_ctx` to be set appropriately

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Ack

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: See comment for removing context

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: See comment for removing context

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: See comment for removing context

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: See comment for removing context

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: See comment for removing context

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Applied

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Tested this with Filestore and it works correctly. UnstructuredFileLoader requires a local file in the filesystem.

Shelnutt2: Sorry that was a typo I meant FileIOLoader. Since the FileStore doesn't fully implemented the python FileIO spec yet was wondering if it was enough. Glad to hear it is!

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
Shelnutt2: Thanks for the explanation!

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: Added tests for this

apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
NikolaosPapailiou: This function is not currently utilized for s3. We could reuse the same import pattern for libraries that the readers require but are not a dependency of the vector search module.

jparismorgan (approved):

PR #285, opened 2024-03-18
Type erased Vamana index

jparismorgan:
### What
Updates the Python Vamana index to be able to ingest vectors and query for them. 

### Testing
* Existing unit tests pass
* New Python and C++ unit tests pass

### TODO
* Right now the query only returns the first result successfully. This first follow-up is to fix this and then check that we're working using SIFT.
* We also don't return external IDs correctly. Will fix that in a follow-up.


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -1457,6 +1460,121 @@ def ingest_flat(
...
+        logger = setup(config, verbose)
+        with tiledb.scope_ctx(ctx_or_config=config):
+            updated_ids = read_updated_ids(
+                updates_uri=updates_uri,
+                config=config,
+                verbose=verbose,
+                trace_id=trace_id,
+            )
+            group = tiledb.Group(index_group_uri)
+            parts_array_uri = group[PARTS_ARRAY_NAME].uri
```
NikolaosPapailiou, line 1491: I think we should use the temp input arrays for this `INPUT_VECTORS_ARRAY_NAME, EXTERNAL_IDS_ARRAY_NAME`

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2520,21 +2688,30 @@ def consolidate_and_vacuum(
...
-            index_timestamp = int(time.time() * 1000)
-        ingestion_timestamps.append(index_timestamp)
-        base_sizes.append(temp_size)
-        partition_history.append(partitions)
-        group.meta["partition_history"] = json.dumps(partition_history)
-        group.meta["base_sizes"] = json.dumps(base_sizes)
-        group.meta["ingestion_timestamps"] = json.dumps(ingestion_timestamps)
-        group.close()
+        
+        if index_type != "VAMANA":
```
NikolaosPapailiou, line 2692: We can introduce some util functions to check if an algorithm is `type-erased` in order to perform the logic here and above?

jparismorgan: Thanks, good call, updated.

apis/python/src/tiledb/vector_search/ingestion.py
jparismorgan: Thanks, good suggestion. Talked offline and then updated to use the same `"PARTIAL_WRITE_ARRAY_DIR": "temp_data",` which ivf_flat uses to write temp data because using `INPUT_VECTORS_ARRAY_NAME, EXTERNAL_IDS_ARRAY_NAME` is tricky with how we already create and write to it if `input_vectors` or `external_ids` are passed.

NikolaosPapailiou (approved):

PR #286, opened 2024-03-18
Add a type-erased Vamana index (which only supports creation)

jparismorgan:
### What
Here we add a type-erased Vamana index which only supports creation. In a follow-up I will add support for querying and ingestion. I do this because this PR already has a good amount of interesting changes, so I want to keep it as simple as possible. Changes included:
* Add `vamana_index.py` which in `create()` will create an empty Vamana index and write it to the specific URI
* Fix crashes in `write_matrix()` and `write_vector()` that would happen during the `index/vamana_index.h@write_index()` call when we try to write an empty vector.
* Update the Vamana index (i.e. `vamana_group.h`) to use `parts_array_name` to store feature vectors instead of a custom `feature_vectors_array_name`. This matches what FLAT does and will make the Python code simpler.
  * We also delete some tests from `unit_vamana_group.cc` because of this - these fail because the previously created indexes use `feature_vectors_array_name`.
* Update the Vamana index (i.e. `index/vamana_index.h`) to store IDs in `ids_array_name`, and update to `ColMajorMatrixWithIds<feature_type, id_type> feature_vectors_;` along with that so we can store the IDs in the index.

### Testing
* Existing unit tests pass.
* New unit tests pass.


jparismorgan:
Opening up this PR for review before `Python API CI` is passing b/c that is a separate issue which is being discussed here: https://tiledb.slack.com/archives/C06H3JS58AD/p1710935849456319?thread_ts=1710777709.344369&cid=C06H3JS58AD

I will wait until it passes to merge this, but we can still get this reviewed in the meantime.

apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Open a Vamana index
+
+    Parameters
+    ----------
+    uri: str
+        URI of the index
+    config: Optional[Mapping[str, Any]]
+        config dictionary, defaults to None
+    memory_budget: int
```
NikolaosPapailiou, line 29: nit: remove `memory_budget`

apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -0,0 +1,154 @@
...
+        config: Optional[Mapping[str, Any]] = None,
+        timestamp=None,
+        **kwargs,
+    ):
+        super().__init__(uri=uri, config=config, timestamp=timestamp)
+        self.index_type = INDEX_TYPE
+        self.index = vspy.IndexVamana(vspy.Ctx(config), uri)
+        self.db_uri = self.group[
+            storage_formats[self.storage_version]["PARTS_ARRAY_NAME"]
+            + self.index_version
```
NikolaosPapailiou, line 45: I don't think that `index_version` is needed for this new type of index. It is used only for backwards compatibility for the existing indexes.

apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            self.size = self.base_size
+
+    def get_dimensions(self):
+        return self.dimensions
+
+    def query_internal(
+        self,
+        queries: np.ndarray,
+        k: int = 10,
```
NikolaosPapailiou, line 72: Lets add the extra query args only when we have an actual distributed query implementation.

apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -0,0 +1,154 @@
...
+            adjacency_row_index_type=np.dtype(adjacency_row_index_type).name, 
+            dimension=dimensions,
+        )
+        empty_vector = vspy.FeatureVectorArray(
+            dimensions, 
+            0, 
+            np.dtype(vector_type).name, 
+            np.dtype(id_type).name
+            )
+        index.train(empty_vector)
```
NikolaosPapailiou, line 151: I think we should try to simplify this code by providing one method in the C++ API that creates an empty index.

apis/python/src/tiledb/vector_search/vamana_index.py
jparismorgan: Thanks, done.

apis/python/src/tiledb/vector_search/vamana_index.py
jparismorgan: Thanks, done.

apis/python/src/tiledb/vector_search/vamana_index.py
jparismorgan: Thanks, done.

apis/python/src/tiledb/vector_search/vamana_index.py
jparismorgan: Agree, though for now I'd prefer to leave as-is just for now - I don't yet have everything working end to end and it's faster to make changes in Python than in native code. Added a TODO and will update once it's fully working, if okay with you?

NikolaosPapailiou (approved):

PR #291, opened 2024-03-22
Bump up tiledb version to 2.21.1

dudoslav:
Bump up:

TileDB-Py version to 0.27.0
TileDB version to 2.21.1

NOTE: I had to modify `DownloadPrebuiltTileDB.cmake` since we are still using superbuild. Also source distribution needs its own logic added.

This should be release: `2.0`


ihnorton (approved):
(FWIW I thought we were going to pull the `DownloadPrebuildTiledb.cmake` from the release bundle, but up to you if you prefer to vendor it)

src/include/test/unit_api_ivf_flat_index.cc
```diff
@@ -282,7 +282,7 @@ TEST_CASE(
     CHECK(nt == nv);
     auto recall = ((double)intersections_a) / ((double)nt * k_nn);
     if (nprobe == 32) {
-      CHECK(recall == 1.0);
+      CHECK(recall >= .999);
```
ihnorton, line 285: @lums658 @jparismorgan please confirm if this is ok. The rest of the higher-precision comparisons in this file are like this. We saw a failure here in [this run](https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/8438568931/job/23111048191?pr=291):
```
9: /home/runner/work/TileDB-Vector-Search/TileDB-Vector-Search/src/include/test/unit_api_ivf_flat_index.cc:285: FAILED:
9:   CHECK( recall == 1.0 )
9: with expansion:
9:   0.999 == 1.0
9: with message:
```

ihnorton (changes requested):
Needs approval of the CI-fix commit I just added, before merging.

jparismorgan (approved):

ihnorton (approved):

PR #296, opened 2024-03-27
Fix source distribution config.h in superbuild

dudoslav:
Source distribution did not include `config.h` so the local build always failed.

In order to fix this I enabled git querying in superbuild so that the `config.h` will be produced before sdist packaging.


dudoslav:
Tested on my local fork: https://github.com/dudoslav/TileDB-Vector-Search/actions/runs/8456266912/job/23165649026

```
(base) taco:removeme dudoslav$ ls tiledb_vector_search-0.0.322/src/
benchmarks  cmake  CMakeLists.txt  config.h  include  README.md  src
```

jparismorgan (approved):

NikolaosPapailiou (approved):

PR #308, opened 2024-04-04
Support access_credentials_name for BATCH tasks

Shelnutt2:
TileDB Cloud supports using certain types of credentials like AWS IAM roles, Azure SAS tokens as part of a batch task graph. These credentials can provide direct access to object stores which is useful in some cases, especially during ingestion. This PR extends the existing support for ingestion by adding the new parameter `acn` which is passed to batch tasks.


ihnorton (approved):
I'm ok with this, but IMO we should look at refactoring soon so to avoid threading complex arguments everywhere.

Shelnutt2:
> I'm ok with this, but IMO we should look at refactoring soon so to avoid threading complex arguments everywhere.

Agreed.

PR #311, opened 2024-04-05
Add support for external ids with Vamana type-erased

jparismorgan:
### What
Before we would pass in the external IDs in Python to `FeatureVectorArray`, i.e.:
```
index = vspy.IndexVamana(ctx, index_group_uri)
data = vspy.FeatureVectorArray(ctx, parts_array_uri, ids_array_uri)
index.train(data)
index.add(data)
index.write_index(ctx, index_group_uri)
```
But then in C++ we would just ignore these IDs. Here we update to read those and map back to external ids during `query()`.

### Testing
* Adds new unit tests which check that external IDs are returned correctly.
* Existing unit tests pass.

### TODO
In the course of this I noticed that we currently only support `uint64` for IDs from Python. I added a note about this and will fix in a follow-up.


NikolaosPapailiou (commented):
As I understand it, the implementation already had a concept of IDs but these IDs were auto-generated during train. Can we instead of adding one more ID mapping use the external IDs directly during the graph creation algorithm? This approach requires extra space and an extra ID translation so I would like to avoid it. Maybe this is a good solution if we have different type of IDs but for now we just have `uint64`

jparismorgan:
> As I understand it, the implementation already had a concept of IDs but these IDs were auto-generated during train. Can we instead of adding one more ID mapping use the external IDs directly during the graph creation algorithm? This approach requires extra space and an extra ID translation so I would like to avoid it. Maybe this is a good solution if we have different type of IDs but for now we just have `uint64`

So the IDs in the current implementation are just indexes. In my discussions with @lums658 we discussed it being complicated (and perhaps problematic) to do this mapping within the graph itself, so the suggestion was to do a mapping afterwards.

Also, if doing the mapping in the graph will make it harder to support other types of IDs, I don't  think we should do it. As in C++ we do actually support a few types of IDs already (`uint32` and `uint64`).

src/include/index/vamana_index.h
```diff
@@ -95,7 +96,8 @@ auto greedy_search(
     auto&& query,
     size_t k_nn,
     size_t L,
-    Distance distance = Distance{}) {
+    Distance distance = Distance{},
+    bool convert_to_db_ids = false) {
```
NikolaosPapailiou, line 100: Why do we need `convert_to_db_ids`? The queries should always return the external_ids or the default ids if external ids were not provided

NikolaosPapailiou (approved):

PR #312, opened 2024-04-05
Support access credentials name for vamana ingestion

Shelnutt2:
This adds the kwargs argument used for other ingestors to vamana to support batch task graph credentials.


NikolaosPapailiou (approved):

jparismorgan (approved):
Thanks for fixing this, missed this in the merge.

PR #314, opened 2024-04-05
Restructure Matrix read operations

NikolaosPapailiou:
Restructure Matrix read operations to handle incomplete queries and batching

### DIFF
--- a/src/include/detail/linalg/tdb_matrix.h
+++ b/src/include/detail/linalg/tdb_matrix.h
@@ -351,20 +351,31 @@ class tdbBlockedMatrix : public MatrixBase {
 
     auto layout_order = schema_.cell_order();
 
-    // Create a query
+    // Read TileDB data
+    size_t total_size = elements_to_load * dimension;
+    size_t offset = 0;
     tiledb::Query query(ctx_, *array_);
-    query.set_subarray(subarray)
-        .set_layout(layout_order)
-        .set_data_buffer(attr_name, this->data(), elements_to_load * dimension);
-    tiledb_helpers::submit_query(tdb_func__, uri_, query);
+    query.set_subarray(subarray).set_layout(layout_order);
+    tiledb::Query::Status status;
+    do {
+      // Submit query and get status
+      size_t request_size = READ_BATCH_SIZE;
+      if (offset + READ_BATCH_SIZE > total_size) {
+        request_size = total_size - offset;
+      }
+      query.set_data_buffer(attr_name, this->data() + offset, request_size);
+      query.submit();
+      status = query.query_status();
+
+      auto num_results = query.result_buffer_elements()[attr_name].second;
+      offset += num_results;
+    } while (status == tiledb::Query::Status::INCOMPLETE);
+    // Handle error
+    if (status == tiledb::Query::Status::FAILED) {
+      throw std::runtime_error("Error in reading");
+    }
     _memory_data.insert_entry(
         tdb_func__, elements_to_load * dimension * sizeof(T));
-
-    // @todo Handle incomplete queries.
-    if (tiledb::Query::Status::COMPLETE != query.query_status()) {
-      throw std::runtime_error("Query status is not complete");
-    }
-
     num_loads_++;
     return true;
   }
--- a/src/include/detail/linalg/tdb_matrix_with_ids.h
+++ b/src/include/detail/linalg/tdb_matrix_with_ids.h
@@ -224,19 +224,33 @@ class tdbBlockedMatrixWithIds
 
     auto layout_order = ids_schema_.cell_order();
     this->ids().resize(elements_to_load * dimension);
-    // Create a query
+
+    // Read TileDB data
+    size_t total_size = elements_to_load * dimension;
+    size_t offset = 0;
+    auto ptr = this->ids().data();
     tiledb::Query query(this->ctx_, *ids_array_);
-    query.set_subarray(subarray)
-        .set_layout(layout_order)
-        .set_data_buffer(attr_name, this->ids());
-    tiledb_helpers::submit_query(tdb_func__, ids_uri_, query);
+    query.set_subarray(subarray).set_layout(layout_order);
+    tiledb::Query::Status status;
+    do {
+      // Submit query and get status
+      size_t request_size = READ_BATCH_SIZE;
+      if (offset + READ_BATCH_SIZE > total_size) {
+        request_size = total_size - offset;
+      }
+      query.set_data_buffer(attr_name, ptr + offset, request_size);
+      tiledb_helpers::submit_query(tdb_func__, ids_uri_, query);
+      status = query.query_status();
+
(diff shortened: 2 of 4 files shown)


jparismorgan (approved):
Thanks for doing this!

(Optional)(Non-blocking) It would be really quite nice if we could unit test this, as if it breaks right now we'll have no way to know. Maybe we could set `readBatchSize` in the constructor and default to `READ_BATCH_SIZE`, and then in some unit tests set this to a smaller number.

src/include/tdb_defs.h
```diff
@@ -36,6 +36,8 @@
 #include <string>
 #include <tiledb/tiledb>
 
+constexpr size_t READ_BATCH_SIZE = 100000000;
```
Shelnutt2, line 39: Lets make this a configuration option for the user at runtime. Let's also be clear what this value is (bytes?).

src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -224,19 +224,33 @@ class tdbBlockedMatrixWithIds
...
-    tiledb_helpers::submit_query(tdb_func__, ids_uri_, query);
+    query.set_subarray(subarray).set_layout(layout_order);
+    tiledb::Query::Status status;
+    do {
+      // Submit query and get status
+      size_t request_size = READ_BATCH_SIZE;
+      if (offset + READ_BATCH_SIZE > total_size) {
+        request_size = total_size - offset;
+      }
+      query.set_data_buffer(attr_name, ptr + offset, request_size);
```
Shelnutt2, line 241: Should there be a check for ensuring we don't extend past the size of the data buffer?

src/include/detail/linalg/tdb_matrix_with_ids.h
```diff
@@ -224,19 +224,33 @@ class tdbBlockedMatrixWithIds
...
+      tiledb_helpers::submit_query(tdb_func__, ids_uri_, query);
+      status = query.query_status();
+
+      auto num_results = query.result_buffer_elements()[attr_name].second;
+      offset += num_results;
+    } while (status == tiledb::Query::Status::INCOMPLETE);
```
Shelnutt2, line 247: This needs an additional check for incompletes. You need to check the reason for the incomplete and ensure the buffers are large enough. See the example from [docs](https://docs.tiledb.com/main/how-to/arrays/reading-arrays/incomplete-queries).

src/include/detail/linalg/tdb_partitioned_matrix.h
```diff
@@ -559,24 +559,32 @@ class tdbPartitionedMatrix
...
-      // @todo Handle incomplete queries.
-      if (tiledb::Query::Status::COMPLETE != query.query_status()) {
-        throw std::runtime_error("Query status is not complete -- fix me");
+      // Read TileDB data
+      size_t total_size = col_count * dimension;
+      size_t offset = 0;
+      tiledb::Query query(ctx_, *(this->partitioned_vectors_array_));
+      query.set_subarray(subarray).set_layout(layout_order);
+      tiledb::Query::Status status;
+      do {
```
Shelnutt2, line 570: Same questions for this loop as the other one. Ensuring we don't extend past the size of the data buffer and properly handling the incomplete query checks.

Shelnutt2 (changes requested):

src/include/detail/linalg/tdb_matrix_with_ids.h
ihnorton: See here: https://github.com/TileDB-Inc/TileDB-Py/blob/dab850604685c89111a0128729d770ab0650ce24/tiledb/core.cc#L1065-L1069

src/include/detail/linalg/tdb_matrix_with_ids.h
NikolaosPapailiou: Restructured the loop and error handling. Let me know if this addresses your comments

src/include/detail/linalg/tdb_partitioned_matrix.h
NikolaosPapailiou: Restructured the loop and error handling. Let me know if this addresses your comments

src/include/tdb_defs.h
NikolaosPapailiou: Made this configurable and added more clarification about the value. This is expressed in number of array cells.

src/include/tdb_defs.h
```diff
@@ -36,6 +36,28 @@
 #include <string>
 #include <tiledb/tiledb>
 
+// Default batch size for all TileDB read operations.
+// This is expressed in number of array cells read per request.
+constexpr size_t DEFAULT_READ_BATCH_SIZE_CELLS = 100000000;
```
lums658, line 41: I suggest using `100'000'000` for legibility.

src/include/detail/linalg/tdb_matrix.h
```diff
@@ -351,20 +351,44 @@ class tdbBlockedMatrix : public MatrixBase {
...
+    size_t total_size = elements_to_load * dimension;
+    size_t offset = 0;
     tiledb::Query query(ctx_, *array_);
-    query.set_subarray(subarray)
-        .set_layout(layout_order)
-        .set_data_buffer(attr_name, this->data(), elements_to_load * dimension);
-    tiledb_helpers::submit_query(tdb_func__, uri_, query);
+    query.set_subarray(subarray).set_layout(layout_order);
+    tiledb::Query::Status status;
+    do {
```
lums658, line 361: Does this result in incomplete queries whenever `read_batch_size_cells` is less than `total_size`?  I don't think we should batch read unless we actually need to.  That is, we should always try to read `total_size` and then iterate if we don't get `total_size`.  

Having incomplete queries here is a very edge case.  Reading in batches may or may not solve that issue, but it adds latency to the common case.

lums658 (changes requested):
It looks like the reads are reading in batches until the total size of the array is reached.  I don't think this is a good way to do this.  The initial request should the `total_size` we want to read, as that will complete in the vast vast majority of cases.  If it doesn't, then we should continue the query and get more.  

There is no reason to batch things in this way -- that is a pattern for a different use case.

Especially if there is a fixed overhead per query, we want to minimize the number of actual queries submitted.

src/include/detail/linalg/tdb_matrix.h
lums658: One rule in software is "don't pay for what you don't need."  It is true that there is only a 5% penalty with this particular set of assumptions.  However, I expect the server is going to deal with its memory footprint in its own way, which is why we might get incomplete reads.  I would say just request what is needed and let the server say how much memory it wants to devote to that or not.

Do other apps try to be parsimonious this way?

I guess, on the other hand, since this is configurable, we can always just make it really large.

lums658 (commented):
See in-line comment.

PR #316, opened 2024-04-08
Fix local ingestion by passing a dummy default namespace

NikolaosPapailiou:
Tested and verified locally with empty and invalid token.


jparismorgan (approved):
Seems good to me, but you may want to wait for @Shelnutt2 to review as well.

PR #317, opened 2024-04-08
Fix type-erased Vamana `query()`

jparismorgan:
### What
When `opt_l < k`, we only return `opt_l` vectors instead of `k` vectors. So here we update the default value of `opt_l` to be based on the C++ implementation and then throw if it is not. With this we can now query with Vamana.

### Testing
* Unit tests pass
* Adds a new unit tests

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -130,6 +130,9 @@ class Index:
         self.thread_executor = futures.ThreadPoolExecutor()
         self.has_updates = self.check_has_updates()
 
+    def is_type_erased_index(self):
+        return self.index_type == "VAMANA"
+
     def query(self, queries: np.ndarray, k, **kwargs):
         if queries.ndim != 2:
             raise TypeError(
@@ -185,7 +188,9 @@ class Index:
                 if (
                     internal_results_d[query_id, res_id] == 0
                     and internal_results_i[query_id, res_id] == 0
+                    and not self.is_type_erased_index()
                 ):
+                    # NOTE(paris): This is skipped for type-erased modules because they will explicitly handle this case internally. It is valid to have an ID of 0 with a score of 0.
                     internal_results_d[query_id, res_id] = MAX_FLOAT_32
                     internal_results_i[query_id, res_id] = MAX_UINT64
                 res_id += 1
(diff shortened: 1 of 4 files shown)


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -185,7 +188,9 @@ def query(self, queries: np.ndarray, k, **kwargs):
                 if (
                     internal_results_d[query_id, res_id] == 0
                     and internal_results_i[query_id, res_id] == 0
+                    and not self.is_type_erased_index()
                 ):
+                    # NOTE(paris): This is skipped for type-erased modules because is is valid to have an ID of 0 with a score of 0.
                     internal_results_d[query_id, res_id] = MAX_FLOAT_32
                     internal_results_i[query_id, res_id] = MAX_UINT64
```
jparismorgan, line 195: I  disabled this check which will throw out query results if the ID is 0 and the score is 0. @NikolaosPapailiou or @lums658 , do you know why we had it? Should we perhaps also remove it for non-type-erased code?
```
if (internal_results_d[query_id, res_id] == 0 and internal_results_i[query_id, res_id] == 0):
    internal_results_d[query_id, res_id] = MAX_FLOAT_32
    internal_results_i[query_id, res_id] = MAX_UINT64
```

NikolaosPapailiou: The convention for all our query implementations is that empty results should be represented as `MAX_UINT64, MAX_FLOAT_32` I think we should be consistent here and fix it on the C++ side rather than patching it here

jparismorgan: Yes, that's why I removed the check here. I don't understand why we have it because in the type-erased code we use the convention you describe. 

I was just asking if I could also remove it for the non-type-erased case, i.e. if you knew why we had this check?

NikolaosPapailiou: Thanks sorry for the misunderstanding, yes at some point our queries returned 0 for empty results. This was added for backwards compatibility. I think we can remove it now

NikolaosPapailiou (approved):

apis/python/src/tiledb/vector_search/index.py
jparismorgan: Sounds good, I'll remove it in a follow-up to keep the changes self-contained.

jparismorgan: Removed here: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/319

PR #322, opened 2024-04-09
Add needs dependency on backwards_compatibility_data

dudoslav:
This PR adds dependency to upload step to wait for generate backwards compatibility data before actually uploading pip wheels to pypi. This way we can decrease number of micro version bumps if generate backwards compatibility data step fails.

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -86,8 +86,6 @@ jobs:
 
   build_wheels:
     name: Wheel ${{ matrix.buildplat[0] }}-${{ matrix.buildplat[1] }}-${{ matrix.python }}
-    # TODO(paris): Add this back once generate_backwards_compatibility_data is confirmed to work.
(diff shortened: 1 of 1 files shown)


jparismorgan (approved):
Thanks! Could also remove these lines if you wanted, had a comment about doing this in the wrong spot: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/322/files#diff-14b51e9e6321caf5d23b2253f2415daf475fdaa417c0aa8fe896c3a460f2d023R89-R90

```
 name: Wheel ${{ matrix.buildplat[0] }}-${{ matrix.buildplat[1] }}-${{ matrix.python }}
    # TODO(paris): Add this back once generate_backwards_compatibility_data is confirmed to work.
    # needs: generate_backwards_compatibility_data
```

jparismorgan:
Actually - can we hold off on merging this for now?

PR #327, opened 2024-04-11
Only pass TILEDB_REST_TOKEN to test_cloud during CI

NikolaosPapailiou:
We don't set `TILEDB_REST_TOKEN` directly in the CI environment. Instead we use an extra environment variable `TILEDB_CLOUD_HELPER_VAR` and only use it during `test_cloud` testing.


jparismorgan (approved):

NikolaosPapailiou:
Only pass TILEDB_REST_TOKEN to test_cloud and Notebook tests during CI

PR #328, opened 2024-04-11
Add option to import OpenAIEmbeddings from langchain_openai

NikolaosPapailiou:
`OpenAIEmbeddings` was removed from the main langchain repo and needs to be imported from `langchain_openai` for newer langchain versions.

### DIFF
--- a/apis/python/src/tiledb/vector_search/embeddings/langchain_embedding.py
+++ b/apis/python/src/tiledb/vector_search/embeddings/langchain_embedding.py
@@ -16,16 +16,19 @@ class LangChainEmbedding:
         dimensions: int,
         embedding_class: str = "OpenAIEmbeddings",
         embedding_kwargs: Optional[Dict] = None,
+        use_langchain_openai: bool = True,
     ):
         self.dim_num = dimensions
         self.embedding_class = embedding_class
         self.embedding_kwargs = embedding_kwargs
+        self.use_langchain_openai = use_langchain_openai
 
     def init_kwargs(self) -> Dict:
         return {
             "dimensions": self.dim_num,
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledb/vector_search/embeddings/langchain_embedding.py
```diff
@@ -37,7 +40,10 @@ def vector_type(self) -> np.dtype:
     def load(self) -> None:
         import importlib
 
-        embeddings_module = importlib.import_module("langchain.embeddings")
+        if self.use_langchain_openai:
```
Shelnutt2, line 43: Instead of having the user pass a parameter can we do an import and try/except to handle this automatically?

See example at: https://github.com/TileDB-Inc/TileDB-ML/blob/effa1c1c7a80026540ce157d4d01c47207ce5ad7/tests/models/test_tensorflow_keras_models.py#L19

NikolaosPapailiou: At the moment this is just a warning, will also need an option to load embeddings from `langchain_community`. Should I just load from the new modules?

ihnorton: I'm fine with setting minimum versions or additional package requirements to keep this as simple as possible.

NikolaosPapailiou: Applied Seth's suggestion with a preference to first try to import from `langchain_openai`

jparismorgan (approved):
Looks good to me, though you may also want to wait on Seth's review.

Shelnutt2 (approved):

PR #329, opened 2024-04-12
Fix `consolidate_updates` error for `tiledb://` URIs.

NikolaosPapailiou:
Fix `consolidate_updates` error for `tiledb://` URIs.

`consolidate_updates` performs `timestamp` consolidation of all updates between consequent ingestions.
This caused failures for file indexing when appending to an existing `tiledb://` index.

This is an optional performance enhancement. We skip it for remote arrays as consolidation of remote arrays currently only supports modes `fragment_meta, commits, metadata`.

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -395,19 +395,23 @@ class Index:
         """
         from tiledb.vector_search.ingestion import ingest
 
-        fragments_info = tiledb.array_fragments(
-            self.updates_array_uri, ctx=tiledb.Ctx(self.config)
-        )
-        max_timestamp = self.base_array_timestamp
-        for fragment_info in fragments_info:
-            if fragment_info.timestamp_range[1] > max_timestamp:
-                max_timestamp = fragment_info.timestamp_range[1]
-        max_timestamp += 1
-        conf = tiledb.Config(self.config)
-        conf["sm.consolidation.timestamp_start"] = self.latest_ingestion_timestamp
-        conf["sm.consolidation.timestamp_end"] = max_timestamp
-        tiledb.consolidate(self.updates_array_uri, config=conf)
-        tiledb.vacuum(self.updates_array_uri, config=conf)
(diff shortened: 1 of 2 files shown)


jparismorgan (approved):
Looks good, but where did you see this error? Can we add a new unit test that will make sure this is working?

NikolaosPapailiou:
> Looks good, but where did you see this error? Can we add a new unit test that will make sure this is working?

Encountered this error while testing the file indexing workflow and adding files to an existing index.

Added a test in `test_cloud` to cover `update_consolidation`

jparismorgan (approved):

NikolaosPapailiou:
Fix `consolidate_updates` error for `tiledb://` URIs.

`consolidate_updates` performs `timestamp` consolidation of all updates between consequent ingestions.
This caused failures for file indexing when appending to an existing `tiledb://` index.

This is an optional performance enhancement. We skip it for remote arrays as consolidation of remote arrays currently only supports modes `fragment_meta, commits, metadata`.

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -403,11 +403,15 @@ def consolidate_updates(self, retrain_index: bool = False, **kwargs):
...
         max_timestamp += 1
-        conf = tiledb.Config(self.config)
-        conf["sm.consolidation.timestamp_start"] = self.latest_ingestion_timestamp
-        conf["sm.consolidation.timestamp_end"] = max_timestamp
-        tiledb.consolidate(self.updates_array_uri, config=conf)
-        tiledb.vacuum(self.updates_array_uri, config=conf)
+        # Consolidate all updates since the previous ingestion_timestamp.
+        # This is a performance optimization. We skip this for remote arrays as consolidation
+        # of remote arrays currently only supports modes `fragment_meta, commits, metadata`.
+        if not self.updates_array_uri.startswith("tiledb://"):
```
ihnorton, line 409: We should make this conditional on the libtiledb version < 2.24 or add some other reminder to check this in a few releases.

PR #330, opened 2024-04-12
Fix TileDB URIs with type-erased indexes, add Vamana test to test_cloud.py

jparismorgan:
### What
* Fixes two issues when using `tiledb://` URIs with type-erased indexes
  1. You cannot use `VFS` with these or it will throw, so instead update to use `bool exists(const tiledb::Context& ctx) const { return tiledb::Object::object(ctx, group_uri_).type() == tiledb::Object::Type::Group; }`
  2. Use relative paths for objects in groups, just like we did with Python here: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/193
* Adds a Vamana test to `test_cloud.py` which exercises this code path (and is how I found these bugs).

### Testing
* Existing tests pass.
* Adds a new test.

### DIFF
--- a/apis/python/test/test_cloud.py
+++ b/apis/python/test/test_cloud.py
@@ -26,6 +26,7 @@ class CloudTests(unittest.TestCase):
         rand_name = random_name("vector_search")
         test_path = f"tiledb://{namespace}/{storage_path}/{rand_name}"
         cls.flat_index_uri = f"{test_path}/test_flat_array"
+        cls.vamana_index_uri = f"{test_path}/vamana_array"
         cls.ivf_flat_index_uri = f"{test_path}/test_ivf_flat_array"
         cls.ivf_flat_random_sampling_index_uri = (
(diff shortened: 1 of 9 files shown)


NikolaosPapailiou (approved):

apis/python/test/test_cloud.py
```diff
@@ -67,6 +69,31 @@ def test_cloud_flat(self):
...
+            index_type="VAMANA",
+            index_uri=index_uri,
+            source_uri=source_uri,
+            input_vectors_per_work_item=5000,
+            config=tiledb.cloud.Config().dict(),
+            # TODO(paris): Fix and then change to Mode.BATCH.
+            mode=Mode.LOCAL,
+        )
+
+        # TODO(paris): Fix error from loading this URI and then re-enable, and add the rest of the test.
```
ihnorton, line 93: Do we have a story for this? We should also test a simple query for completeness.

PR #332, opened 2024-04-15
Fix tiledb direct URIs in C++ group's

jparismorgan:
### What
Right now doing this will fail on `vs.vamana_index.VamanaIndex(uri=tiledb_index_uri)`:
```
vs.ingest(
            index_type="VAMANA",
            index_uri=index_uri,
            source_uri=source_uri,
            input_vectors_per_work_item=5000,
            config=tiledb.cloud.Config().dict(),
            # TODO(paris): Fix and then change to Mode.BATCH.
            mode=Mode.LOCAL,
        )

        tiledb_index_uri = groups.info(index_uri).tiledb_uri
        vs.vamana_index.VamanaIndex(uri=tiledb_index_uri)
```
This is because `tiledb_index_uri` will be of the form `tiledb://foo/edc4656a-3f45-43a1-8ee5-fa692a015c53`, not `TODO`, so just doing `uri + "/" + array_in_group_name` will not work. Instead we need to do what Python does, which is get the actual array URIs:
```
self.db_uri = self.group[storage_formats[self.storage_version]["PARTS_ARRAY_NAME"]].uri
```
So here we update C++ to do that.

### Testing
* Now the Python code above runs.
* Existing unit tests pass.


jparismorgan:
TODO(paris): I will look at updating this to open a read group instead of using a map to store the name -> uri mappings.

src/include/index/index_group.h
```diff
@@ -140,29 +135,12 @@ class base_index_group {
       valid_array_keys_.insert(array_key);
       valid_array_names_.insert(array_name);
       array_key_to_array_name_[array_key] = array_name;
+      array_name_to_uri_[array_name] =
```
NikolaosPapailiou, line 138: Why do we need this? We are supposed to get the values from the group after this method is executed

jparismorgan: Yep I have not yet updated the PR with that change, this is still the original approach

jparismorgan: @NikolaosPapailiou okay updated now if you want to review!

src/include/index/index_group.h
```diff
@@ -176,13 +147,13 @@ class base_index_group {
       throw std::runtime_error(
           "Group uri " + std::string(group_uri_) + " does not exist.");
     }
-    auto read_group = tiledb::Group(cached_ctx_, group_uri_, TILEDB_READ, cfg);
+    read_group_ = new tiledb::Group(cached_ctx_, group_uri_, TILEDB_READ);
```
NikolaosPapailiou, line 150: If we use this style of Group creation we need to figure out a way to `close` and `delete` the group

jparismorgan: Won't this happen automatically when the base group is destructed and we trigger the read_group_ destructor?

jparismorgan: Okay no I was correct initially, and you're correct, it will not get deleted. Got confused by the comment by Andrew, but it's a raw pointer, so there will be no automatic deletion when it goes out of scope.

NikolaosPapailiou (approved):

src/include/index/index_group.h
jparismorgan: Talked offline - will go back to the map and update to a group after fixing the bug in `cached_ctx_`. See full context here: https://tiledb.slack.com/archives/C0537B4V7Q8/p1713447475923239?thread_ts=1713354838.545549&cid=C0537B4V7Q8

lums658: In the Slack discussion I think we were talking about two different things.  Sorry.  I was talking about `read_group_` as a smart pointer, which was how the discussion thread started.

A raw pointer is non-owning (per C++ Core Guidelines R.3).  I don't know if it is an org requirement, but it should be.  Else for the `new` you have to do a `delete` etc.

jparismorgan: Ah I really got tricked there haha. I was like "well I definitely didn't think this, but I know Andrew knows C++ 😅".

PR #338, opened 2024-04-18
Run Python ingestion tests on all indexes

jparismorgan:
### What
Right now the IVF Flat python tests in `test_ingestion.py` are very thorough. Rather than re-writing these for Vamana (and hypothetically Flat, as we don't test it very well), we just loop through all indexes in these tests.

### Testing
Unit tests pass.


NikolaosPapailiou (approved):
Thanks for adding this! It makes the test more consistent for all implementations.

PR #340, opened 2024-04-19
Remove `std::reference_wrapper` around `tiledb::Context`

jparismorgan:
### What

@lums658 did a great investigation and found that we don't need to be using `std::reference_wrapper` around `tiledb::Context`, as `tiledb::Context` uses a shared pointer under the hood anyways so is fine to copy:

Screenshot 2024-04-19 at 6 13 25 PM

Here we just remove `std::reference_wrapper` around `tiledb::Context`, and in a follow-up I'll update https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/332 to use a `unique_ptr` around a `tiledb::Group` which should work now that the context won't go out of scope.

### Testing
* Existing tests pass

### DIFF
--- a/src/include/detail/linalg/tdb_matrix.h
+++ b/src/include/detail/linalg/tdb_matrix.h
@@ -80,7 +80,7 @@ class tdbBlockedMatrix : public MatrixBase {
 
   log_timer constructor_timer{"tdbBlockedMatrix constructor"};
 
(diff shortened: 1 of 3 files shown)


lums658 (approved):
Looks good to me.

NikolaosPapailiou (approved):

PR #341, opened 2024-04-22
Add support for `ingest()` with `index_timestamp` for type-erased indexes

jparismorgan:
### What
Here we add support for `index_timestamp` to the Vamana index.

To get this to work we changed to not automatically initialize the metadata with `ingestion_timestamps = [0]`, `base_sizes = [0]`, etc. This is because in Python during `ingest()` we will 

1. Call `write_index()` with an empty `FeatureVectorArray` 
2. Call `write_index()` with the actual vectors in `FeatureVectorArray`

In **1.** we will call with `index_timestamp = 0`, and in **2.** we call with `index_timestamp = index_timestamp` (which will default to the current timestamp if `index_timestamp = None`). If we hadn't removed the default of `ingestion_timestamps = [0]`, in **1.** we would fail because we'd be writing with `timestamp = 0`, but we would already have `timestamp = 0` in metadata (and you cannot have two equivalent timestamps saved in metadata).

### Testing
* Existing tests pass
* More `test_ingestion.py` tests pass

### TODO
* In a follow-up I'll also add support for other indexes, but just starting with Vamana to make the PR easier to review.
* The `test_ingestion.py` tests do not all pass yet, so will continue working to get more to pass.

### DIFF
--- a/apis/python/src/tiledb/vector_search/vamana_index.py
+++ b/apis/python/src/tiledb/vector_search/vamana_index.py
@@ -121,5 +121,5 @@ def create(
     )
     index.train(empty_vector)
     index.add(empty_vector)
-    index.write_index(ctx, uri, storage_version)
+    index.write_index(ctx, uri, 0, storage_version)
     return VamanaIndex(uri=uri, config=config, memory_budget=1000000)
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -90,7 +90,7 @@ def ingest(
     updates_uri: str
         Updates
     index_timestamp: int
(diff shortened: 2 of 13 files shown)


apis/python/src/tiledb/vector_search/vamana_index.py
NikolaosPapailiou, line 1: I think we should also change the `__init__` method to open the index at a specified timestamp. 
Feel free to do this in a followup PR.

NikolaosPapailiou (approved):

apis/python/src/tiledb/vector_search/vamana_index.py
jparismorgan: Yep should have mentioned that explicitly, that's why some of the tests are still failing. But just wanted to keep the PR easier to review.

PR #343, opened 2024-04-24
Fix non-contiguous queries when using Vamana

jparismorgan:
### What
Previously if we had `queries` loaded with `queries = load_fvecs(queries_uri)`, they would not be C contiguous (i.e. row major) or F contiguous (i.e. col major):
```
Code:
a = load_fvecs(siftsmall_query_file)[0:99]
print('a', a)
print('a.shape', a.shape)
print('a.flags', a.flags)
a = np.transpose(a)
print('a', a)
print('a.shape', a.shape)
print('a.flags', a.flags)

Output:
a [[ 1.  3. 11. ... 42. 48. 11.]
 [40. 25. 11. ...  3. 19. 13.]
 [28.  4.  3. ...  2. 54. 47.]
 ...
 [13.  7.  4. ... 24. 43. 30.]
 [14.  8.  0. ... 32. 14. 15.]
 [11. 48. 10. ...  0.  0.  0.]]
a.shape (99, 128)
a.flags   C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False

a [[ 1. 40. 28. ... 13. 14. 11.]
 [ 3. 25.  4. ...  7.  8. 48.]
 [11. 11.  3. ...  4.  0. 10.]
 ...
 [42.  3.  2. ... 24. 32.  0.]
 [48. 19. 54. ... 43. 14.  0.]
 [11. 13. 47. ... 30. 15.  0.]]
a.shape (128, 99)
a.flags   C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : False
  ALIGNED : True
  WRITEBACKIFCOPY : False
```
This is because `load_fvecs()` returns a view of the array:
```
apis/python/src/tiledb/vector_search/utils.py

def _load_vecs_t(uri, dtype, ctx_or_config=None):
    ...
            if dtype != np.uint8:
                return raw.view(np.int32).reshape((-1, ndim + 1))[:, 1:].view(dtype)
            else:
                return raw.view(np.uint8).reshape((-1, ndim + 1))[:, 1:].view(dtype)
```
In this case, where we have `C_CONTIGUOUS : False` and `F_CONTIGUOUS : False`,  we would fall back to `TILEDB_ROW_MAJOR` and set rows (i.e. dimensions) to `info.shape[1]` (i.e. 99) and cols (i.e. num vectors) to `info.shape[0]` (i.e. 128). But this is incorrect because as you can see above, once we're transposed `queries`, doing this will lead to us creating a `FeatureVectorArray` with 128 vectors that each have 99 dimensions, instead of 99 vectors with 128 dimensions:
```
apis/python/src/tiledb/vector_search/type_erased_module.cc

py::class_<FeatureVectorArray>(m, "FeatureVectorArray", py::buffer_protocol())
...
def(py::init([](py::array b) {
        /* Request a buffer descriptor from Python */
        py::buffer_info info = b.request();
        ...
        auto v = [&]() {
          auto order = b.flags() & py::array::f_style ? TILEDB_COL_MAJOR :
                                                        TILEDB_ROW_MAJOR;
          if (order == TILEDB_COL_MAJOR) {
            return FeatureVectorArray(info.shape[0], info.shape[1], dtype_str);
          } else {
            return FeatureVectorArray(info.shape[1], info.shape[0], dtype_str);
          }
        }();

        auto data = (uint8_t*)v.data();
        std::memcpy(
            data, (uint8_t*)info.ptr, info.shape[0] * info.shape[1] * sz);

        return v;
      }));
```
To fix this we update the Vamana index `query()` method to rewrite `queries` as `f_contiguous`:
```
if not queries.flags.f_contiguous:
    queries = np.asfortranarray(queries)
```
Note that this looks similar to what the old `array_to_matrix()` code does in accepting the array as `py::array::f_style`:
```
apis/python/src/tiledb/vector_search/module.cc

template <class T>
static void declare_pyarray_to_matrix(
    py::module& m, const std::string& suffix) {
  m.def(
      ("pyarray_copyto_matrix" + suffix).c_str(),
      [](py::array_t<T, py::array::f_style> arr) -> ColMajorMatrix<T> {
        py::buffer_info info = arr.request();
        if (info.ndim != 2)
          throw std::runtime_error("Number of dimensions must be two");
        if (info.format != py::format_descriptor<T>::format())
          throw std::runtime_error("Mismatched buffer format!");

        auto data = std::unique_ptr<T[]>{new T[info.shape[0] * info.shape[1]]};
        std::memcpy(
            data.get(), info.ptr, info.shape[0] * info.shape[1] * sizeof(T));
        auto r =
            ColMajorMatrix<T>(std::move(data), info.shape[0], info.shape[1]);
        return r;
      });
}
```
With this, the `FeatureVectorArray` created from `queries` now works correctly.

### Testing
* Existing tests pass.
* More Vamana ingestion tests now pass.
* I added more `FeatureVectorArray` tests and added TODO's in a few places where things don't behave as expected.

### TODO
There are several TODO's in here about bugs with `FeatureVectorArray`, and I've created a story investigate more: https://app.shortcut.com/tiledb-inc/story/45968


apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -91,8 +91,11 @@ def query_internal(
 
         if queries.ndim == 1:
             queries = np.array([queries])
+        queries = np.transpose(queries)
+        if not queries.flags.f_contiguous:
+            queries = np.asfortranarray(queries)
```
NikolaosPapailiou, line 96: Why not use `.copy(order="C")`?

NikolaosPapailiou (approved):

PR #348, opened 2024-04-26
Fix `Index.delete_index()` so that it uses recursive=True and deletes the full index

jparismorgan:
### What
In a [Vamana CI job](https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/8831344021/job/24246361970?pr=343#step:6:593) I saw it fail consistently when it's installing the jupyter notebook pip packages after running the Python tests because we have disk space after the unit tests:
```
================ 65 passed, 137 warnings in 8725.71s (2:25:25) =================
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        73G   70G  2.8G  97% /                    NOTE(paris): This CI job started at 50G, so we used 20G in the tests
tmpfs           7.9G  172K  7.9G   1% /dev/shm
tmpfs           3.2G  1.1M  3.2G   1% /run
tmpfs           5.0M     0  5.0M   0% /run/lock
/dev/sda15      105M  6.1M   99M   6% /boot/efi
/dev/sdb1        74G  4.1G   66G   6% /mnt
tmpfs           1.6G   12K  1.6G   1% /run/user/1001
```
To fix this I tried adding `Index.delete_index(uri=index_uri, config={})` after each unit test, but I noticed when running locally that the index wasn't actually deleted, i.e. here was an index after it should have been deleted:

Screenshot 2024-04-25 at 4 42 15 PM

To fix this I changed to `group.delete(recursive=True)`, but then I got crashes with:
```
in delete_index
    group.delete(recursive=True)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = array_VAMANA_copied GROUP
|-- input_vectors ARRAY
|-- shuffled_vectors ARRAY
|-- adjacency_ids ARRAY
|-- adjacency_scores ARRAY
|-- adjacency_row_index ARRAY
|-- shuffled_vector_ids ARRAY

recursive = True

    def delete(self, recursive: bool = False):
        """
        Delete a Group. The group needs to be opened in 'm' mode.
    
        :param uri: The URI of the group to delete
        """
>       self._delete_group(self.uri, recursive)
E       tiledb.cc.TileDBError: TileDB internal: Cannot open array; Array does not exist.
```
It turns out these crashes just happened when `input_vectors` was passed to `ingest()`. This is because we would call `tiledb.Array.delete_array(group[INPUT_VECTORS_ARRAY_NAME].uri)`, but we wouldn't remove `INPUT_VECTORS_ARRAY_NAME` from the `group`. So here we update to do that and with it we no longer crash.

### Testing
* Added new tests which check that the size of the index is 0 after calling `Index.delete_index()`.
* Existing tests pass.

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -2505,15 +2505,20 @@ def ingest(
         config: Optional[Mapping[str, Any]] = None,
     ):
         with tiledb.Group(index_group_uri) as group:
+            write_group = tiledb.Group(index_group_uri, "w")
             try:
                 if INPUT_VECTORS_ARRAY_NAME in group:
                     tiledb.Array.delete_array(group[INPUT_VECTORS_ARRAY_NAME].uri)
+                    write_group.remove(INPUT_VECTORS_ARRAY_NAME)
                 if EXTERNAL_IDS_ARRAY_NAME in group:
                     tiledb.Array.delete_array(group[EXTERNAL_IDS_ARRAY_NAME].uri)
+                    write_group.remove(EXTERNAL_IDS_ARRAY_NAME)
             except tiledb.TileDBError as err:
                 message = str(err)
                 if "does not exist" not in message:
                     raise err
+            write_group.close()
+
             modes = ["fragment_meta", "commits", "array_meta"]
             for mode in modes:
                 conf = tiledb.Config(config)
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -2504,8 +2504,8 @@ def ingest(
         index_group_uri: str,
         config: Optional[Mapping[str, Any]] = None,
     ):
-        write_group = tiledb.Group(index_group_uri, "w")
         with tiledb.Group(index_group_uri) as group:
+            write_group = tiledb.Group(index_group_uri, "w")
             try:
                 if INPUT_VECTORS_ARRAY_NAME in group:
                     tiledb.Array.delete_array(group[INPUT_VECTORS_ARRAY_NAME].uri)
@@ -2517,6 +2517,7 @@ def ingest(
                 message = str(err)
                 if "does not exist" not in message:
                     raise err
+            write_group.close()
 
             modes = ["fragment_meta", "commits", "array_meta"]
             for mode in modes:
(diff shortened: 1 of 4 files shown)


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2505,15 +2505,20 @@ def consolidate_and_vacuum(
         config: Optional[Mapping[str, Any]] = None,
     ):
         with tiledb.Group(index_group_uri) as group:
+            write_group = tiledb.Group(index_group_uri, "w")
```
NikolaosPapailiou, line 2508: There are multiple Group open operations in `consolidate_and_vacuum` as well as before calling it, can we open the group once in `m` mode?

NikolaosPapailiou (approved):

apis/python/src/tiledb/vector_search/ingestion.py
jparismorgan: Updated to just have a single `w` within `consolidate_and_vacuum()`, and in a follow-up will look at if we can pass an open group to this method / have less opens in general.

jparismorgan: Hmm, it seems like my update to use a single write group in consolidate_and_vacuum() doesn't work in the cloud tests (https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/8848628336/job/24300692458?pr=348#step:6:378):
```
https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/8848628336/job/24298822778?pr=348#step:6:483 
=========================== short test summary info ============================
FAILED test/test_cloud.py::CloudTests::test_cloud_ivf_flat - tiledb.cc.TileDBError: [TileDB::REST] Error: Error in libcurl PATCH operation: libcurl error message 'CURLE_OK'; HTTP code 500; server response data '{"code":901,"message":"failed to write member changes to core group: asset not found as an array or group: record not found","request_id":"b6c4b094-3755-40e7-9ea2-727d297b19dc"}'.
FAILED test/test_cloud.py::CloudTests::test_cloud_ivf_flat_random_sampling - tiledb.cc.TileDBError: [TileDB::REST] Error: Error in libcurl PATCH operation: libcurl error message 'CURLE_OK'; HTTP code 500; server response data '{"code":901,"message":"failed to write member changes to core group: asset not found as an array or group: record not found","request_id":"634558b7-2c7c-46a9-b1d8-000592eb12fb"}'.
FAILED test/test_cloud.py::CloudTests::test_cloud_vamana - tiledb.cc.TileDBError: [TileDB::REST] Error: Error in libcurl PATCH operation: libcurl error message 'CURLE_OK'; HTTP code 500; server response data '{"code":901,"message":"failed to write member changes to core group: asset not found as an array or group: record not found","request_id":"f3c8dc93-ce04-4129-91cc-a7c107b0f841"}'.
=========== 3 failed, 62 passed, 218 warnings in 1737.60s (0:28:57) ============
Error: Process completed with exit code 1.
```

jparismorgan: Talked offline, and agreed to go back to the original approach and consolidate group opens in a follow-up.

PR #349, opened 2024-04-26
Fix `BATCH` ingestion for Vamana

jparismorgan:


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2183,6 +2175,15 @@ def create_ingestion_dag(
             )
             return d
         elif index_type == "VAMANA":
+            # We must create the partial write array group outside of the UDF b/c tiledb.FilterList cannot be pickled
```
Shelnutt2, line 2178: I don't believe we want this. We need to do creates inside the UDF not in the driver node else it could end up with permission problems requiring the caller to have full access to storage.

jparismorgan: Mmm, so we already do this for IVF Flat outside ([here](https://github.com/TileDB-Inc/TileDB-Vector-Search/blob/main/apis/python/src/tiledb/vector_search/ingestion.py#L683)), and I don't think we have a problem there? 

In general, we do a lot of the array and group creation outside of the UDF - the general pattern is that the groups and arrays are created outside of the UDF and then in the UDF we just do ingestion / centroids computation etc. cc @NikolaosPapailiou in case I've missed some nuance here.

jparismorgan: Got it, thanks. So my proposal is that we move forward with this PR as-is (once I have it passing CI that is) because it doesn't introduce a new thing, and we fix this for both IVF Flat and Vamana together after. Would you be okay with that @Shelnutt2?

jparismorgan: Okay talked to Nikos and have a new approach that leaves it in the UDF! The real issue was that we were implicitly importing `tiledb.FilterList` and so had to pickle it. By explicitly importing it in the UDF and then passing it to `create_partial_write_array_group()` we fix the error.

NikolaosPapailiou (approved):

PR #353, opened 2024-04-29
Update `FeatureVectorArray` to be able to read from URIs at a specified timestamp

jparismorgan:
### What
Here we update `FeatureVectorArray` so that it can read from URIs at a specified timestamp. This will be used to update the type-erased Vamana index so that it can be opened at a specific timestamp.

### Testing
* Adds a new unit test which writes at timestamp 99 and 100 and then validates by default we read at timestamp 100 but we can also explicitly read at timestamp 99.

### DIFF
--- a/src/include/api/feature_vector_array.h
+++ b/src/include/api/feature_vector_array.h
@@ -83,8 +83,13 @@ class FeatureVectorArray {
       const tiledb::Context& ctx,
       const std::string& uri,
       const std::string& ids_uri = "",
-      size_t num_vectors = 0) {
-    auto array = tiledb_helpers::open_array(tdb_func__, ctx, uri, TILEDB_READ);
+      size_t num_vectors = 0,
+      size_t timestamp = 0) {
+    auto temporal_policy =
+        timestamp == 0 ? tiledb::TemporalPolicy() :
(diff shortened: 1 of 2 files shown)


src/include/api/feature_vector_array.h
```diff
@@ -83,8 +83,13 @@ class FeatureVectorArray {
       const tiledb::Context& ctx,
       const std::string& uri,
       const std::string& ids_uri = "",
-      size_t num_vectors = 0) {
-    auto array = tiledb_helpers::open_array(tdb_func__, ctx, uri, TILEDB_READ);
+      size_t num_vectors = 0,
+      size_t timestamp = 0) {
```
NikolaosPapailiou, line 87: Can we support a `TemporalPolicy` argument?

NikolaosPapailiou (approved):

src/include/api/feature_vector_array.h
jparismorgan: Yep I plan to instead of this, but I was planning to do the migration all at once b/c there are many places we need to update. Is that okay with you?

NikolaosPapailiou: sgtm

PR #354, opened 2024-04-29
Vamana type-erased index can be opened at a specific timestamp

jparismorgan:
### What
Previously we couldn't open the Vamana type-erased index at a specific timestamp, now we can. To do this we make several subtle but important changes:
1. We update update so that if we are in our first `ingest()` for an index, we overwrite the metadata created from the first `write_index()` during the second `write_index()` in `ingest_vamana`. This is because if we are creating a new index during ingestion, we will call `write_index()` twice. We overwrite the first `ingestion_timestamps`, `base_sizes` and other lists, otherwise we will end up with `ingestion_timestamps = [0, timestamp]` and `base_sizes = [0, initial size]`, whereas other Python indexes will end up with `ingestion_timestamps = [timestamp]` and `base_sizes = [initial size]`. If we do end up like this then we hit crashes in `index.update()` and `Index.clear_history()` and have to do some small indexing changes. So here we avoid custom Python logic by making sure C++ is set up as Python expects.
2. We don't access `partition_history[i]` in `Index.clear_history()` if we have a type-erased index.
3. We load `FeatureVectorArray` and `IndexVamana` at specific timestamps.

### Testing
* Existing tests pass.
* More of the time travel test in `test_ingestion.py` passes.
* Adds a new C++ test which explicitly tests change # 1. 
* Adds a new C++ test which explicitly tests loading `IndexVamana` at a timestamp (i.e. # 3).


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -490,7 +492,8 @@ def clear_history(
                 if ingestion_timestamp > timestamp:
                     new_ingestion_timestamps.append(ingestion_timestamp)
                     new_base_sizes.append(base_sizes[i])
-                    new_partition_history.append(partition_history[i])
+                    if not is_type_erased_index(index_type):
```
NikolaosPapailiou, line 495: Lets add some comment here to explain why we skip for `type_erased_index`

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -1632,10 +1632,30 @@ def ingest_vamana(
...
         index.add(data)
-        index.write_index(ctx, index_group_uri, index_timestamp)
+        # If we just created the index, we need to overwrite the index metadata. This is because if
+        # we are creating a new index during ingestion, we will call write_index() twice. In this
+        # case we should overwrite the previous ingestion_timestamps and other lists, otherwise we will
+        # end up with ingestion_timestamps = [0, timestamp] and base_sizes = [0, initial size],
+        # whereas other Python indexes will end up with ingestion_timestamps = [timestamp] and
+        # base_sizes = [initial size]. This becomes problematic when we try to modify the index
+        # later (i.e. through update / clear / etc).)
+        overwrite_metadata_list = not arrays_created
```
NikolaosPapailiou, line 1651: This check is a bit subtle and I am not sure it will work consistently for all cases. 

For example, if we do `vamana_index.create` and then `ingest` this will be `false` even if it is the first ingestion. 

Maybe a more consistent check would be to check `base_size==0 && base_ingestion_timestamp==0`. 

I would also suggest to do such checks in C++ and not pass the values from Python in order for the C++ code to be self-contained.

apis/python/src/tiledb/vector_search/index.py
jparismorgan: Sure, done.

apis/python/src/tiledb/vector_search/ingestion.py
jparismorgan: Sure, good suggestion, updated.

NikolaosPapailiou (approved):

PR #356, opened 2024-05-01
Update C++ code to use TemporalPolicy instead of a timestamp

jparismorgan:
### What
Here we update C++ code to use TemporalPolicy instead of a timestamp. The goal is that this is a straight refactor with no logic change / behavior change. This partly addresses SC-39936, though there will be a follow-up to have type-erased Python use these APIs.

### Testing
Existing unit tests pass.

### Why do we implement our own `TemporalPolicy`?
In order to do this we implement our own `TemporalPolicy` instead of using `tiledb::TemporalPolicy`. Note that our version is nearly identical to the core version, so if we change theirs to work in our code then it should be an easy migration over.

This is because the `tiledb::TemporalPolicy`:
1. breaks the copy assignment operator if it is a member variable of a class
2. Does not expose `timestamp_start` and `timestamp_end` as public members which we need (i.e. we need to get it so we can set `index_timestamp_`):
```
src/include/index/index_group.h

void open_for_read() {
    init_for_open();

    if (size(metadata_.ingestion_timestamps_) == 0) {
      throw std::runtime_error("No ingestion timestamps found.");
    }
    if (index_timestamp_ == 0) {
      index_timestamp_ = metadata_.ingestion_timestamps_.back();
    }

    auto timestamp_bound = std::lower_bound(begin(metadata_.ingestion_timestamps_), end(metadata_.ingestion_timestamps_), index_timestamp_);
    if (timestamp_bound == end(metadata_.ingestion_timestamps_)) {
      throw std::runtime_error("Requested read timestamp " + std::to_string(index_timestamp_) + " is beyond " + std::to_string(metadata_.ingestion_timestamps_.back()));
    }
    timetravel_index_ = std::distance(begin(metadata_.ingestion_timestamps_), timestamp_bound);

    // @todo Or index_timestamp_?
    group_timestamp_ = metadata_.ingestion_timestamps_[timetravel_index_];
  }
```
Here you can see `tiledb::TemporalPolicy`:
```
class TemporalPolicy {
  /**
   * Make available the internals of TemporalPolicy to Array.
   */
  friend class Array;

 public:
  TemporalPolicy()
      : timestamp_start_(0)
      , timestamp_end_(UINT64_MAX){};

  TemporalPolicy(const TimeTravelMarker&, uint64_t timestamp)
      : timestamp_start_(0)
      , timestamp_end_(timestamp){};

  TemporalPolicy(
      const TimestampStartEndMarker&,
      uint64_t timestamp_start,
      uint64_t timestamp_end)
      : timestamp_start_(timestamp_start)
      , timestamp_end_(timestamp_end){};

 private:
  inline uint64_t timestamp_start() const {
    return timestamp_start_;
  }

  inline uint64_t timestamp_end() const {
    return timestamp_end_;
  }

  const uint64_t timestamp_start_;
  const uint64_t timestamp_end_;
};
```
and the copy assignment operator error:
```
/Users/parismorgan/repo/Vector-Search-1/src/include/test/unit_vamana_group.cc:202:9: error: object of type 'vamana_index_group<dummy_index>' cannot be assigned because its copy assignment operator is implicitly deleted
      x = vamana_index_group(dummy_index{}, ctx, tmp_uri, TILEDB_WRITE);
        ^
/Users/parismorgan/repo/Vector-Search-1/src/include/index/vamana_group.h:82:28: note: copy assignment operator of 'vamana_index_group<dummy_index>' is implicitly deleted because base class 'base_index_group<vamana_index_group<dummy_index>>' has a deleted copy assignment operator
class vamana_index_group : public base_index_group<vamana_index_group<Index>> {
                           ^
/Users/parismorgan/repo/Vector-Search-1/src/include/index/index_group.h:98:26: note: copy assignment operator of 'base_index_group<vamana_index_group<dummy_index>>' is implicitly deleted because field 'temporal_policy_' has a deleted copy assignment operator
  tiledb::TemporalPolicy temporal_policy_;
                         ^
/Users/parismorgan/repo/TileDB/build-apr18/dist/include/tiledb/array.h:113:18: note: copy assignment operator of 'TemporalPolicy' is implicitly deleted because field 'timestamp_start_' is of const-qualified type 'const uint64_t' (aka 'const unsigned long long')
  const uint64_t timestamp_start_;
                 ^
```


apis/python/test/test_ingestion.py
```diff
@@ -512,6 +512,18 @@ def test_ingestion_with_updates(tmp_path):
             source_uri=os.path.join(dataset_dir, "data.u8bin"),
             partitions=partitions,
         )
+
+        # TODO(paris): Fix Vamana to have same metadata as Python and re-enable.
```
ihnorton, line 516: Do we have a story for this? (and pls cross-link)

ihnorton (approved):
LGTM. Please also file a shortcut to reconsider/update the core API.

apis/python/test/test_ingestion.py
jparismorgan: This is addressed in the follow-up here: #359

PR #359, opened 2024-05-06
Add more C++ and Python temporal policy unit tests

jparismorgan:
### What
* Adds more C++ and Python temporal policy unit tests.
* Fixes the `IndexVamana` constructor in `type-erased-module.cc` to by default use a timestamp of 0. This is a magic value which in the C++ code we will interpret as meaning "we didn't set this, please use the current timestamp". In a follow-up PR I will update this to use `std::optional` so we don't have this magic value of 0.

### Testing
New and old tests pass.


jparismorgan:
NOTE(paris): This PR is not yet ready for review. It builds on a different PR which I'll land first, then I'll pull from `main` here and there will be a smaller diff.

NikolaosPapailiou (approved):

PR #360, opened 2024-05-06
Add `TemporalPolicy` to Python, use `std::optional` instead of `timestamp_end = 0` to mark not set

jparismorgan:
### What
Two big changes which let use support using a `TemporalPolicy` which supports a timestamp start and end (as opposed to just a timestamp end which we support today):
1. We add the C++ `TemporalPolicy` class to Python and pass it to C++ instead of just passing timestamp end.
2. Right now we use a timestamp end of 0 to mark that we have not set the temporal policy. This is tricky and can be confusing to reason about, so here we switch to using `std::optional` to mark that we have not set it explicitly.

### Testing
* Existing unit tests pass.
* The rest of `test_ingestion_with_updates_and_timetravel` now passes.
* Adds several new Python and C++ unit tests which check different scenarios, including that we can load indexes at timestamps before the first ingestion and correctly get back max int / float results.


jparismorgan:
NOTE(paris): This PR is not yet ready for review. It builds on a different PR which I'll land first, then I'll pull from `main` here and there will be a smaller diff.

NikolaosPapailiou (approved):
Thanks for doing this, it makes the API much cleaner!

PR #365, opened 2024-05-09
Fix bug where in Python where if an index was loaded before the earliest ingested data, we'd still load and query data from the future

jparismorgan:
### What
In Python if you ingested with `timestamp = 10` and then loaded an index with `timestamp = 5`, we would still load and query the data which you ingested at `timestamp = 10`, even though that's later then the user asked for.

This also applies to tuples, i.e if you loaded an index with `timestamp = (0, 5)`, we would still load and query the data which you ingested at `timestamp = 10`, even though that's later then the user asked for.

### Testing
* I had to modify `test_ingestion_with_updates` because it was doing `ingest(timestamp=102)`, calling clear_history(timestamp=101)`, and then loading the index at `timestamp=51` and expecting to be able to query with `accuracy = 1.0`. So I changed to expecting `accuracy = 0.0`.
* Adds a big new test that covers `ingest()`, `update()`, `consolidate_updates()`, and `clear_history()` and makes sure we can load an index and query correctly with both int and tuple timestamps.

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -91,39 +91,45 @@ class Index:
                     raise ValueError(
                         "'timestamp' argument expects either int or tuple(start: int, end: int)"
                     )
-                if timestamp[0] is not None:
-                    if timestamp[0] > self.ingestion_timestamps[0]:
-                        self.query_base_array = False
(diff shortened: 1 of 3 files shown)


NikolaosPapailiou (approved):
Thanks for debugging and fixing this!

PR #371, opened 2024-05-13
Add clear_history() to C++, refactor group to allow static function to open a group

jparismorgan:
### What
Here we add C++ support for the `clear_history()` API. In order to do this we refactor `index_group.h` to allow the `clear_history()` static function to open a group. We need this because previously we would have to pass in a group object to `index_group`, now we just pass in the type. This change also has the added benefit of deleting a good amount of code and simplifying things.

### TODO
We don't actually implement the `clear_history()` logic, but we set it up to be called. I'll implement this in a follow-up right after this: #372 

### Testing
* Existing tests pass.
* Can call `clear_history()` from C++.

### DIFF
--- a/src/include/api/vamana_index.h
+++ b/src/include/api/vamana_index.h
@@ -131,30 +131,12 @@ class IndexVamana {
       const tiledb::Context& ctx,
       const URI& group_uri,
       TemporalPolicy temporal_policy = TemporalPolicy{TimeTravel, 0}) {
-    using metadata_element = std::tuple<std::string, void*, tiledb_datatype_t>;
-    std::vector<metadata_element> metadata{
-        {"feature_datatype", &feature_datatype_, TILEDB_UINT32},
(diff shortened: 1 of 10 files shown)


src/include/api/vamana_index.h
```diff
@@ -252,6 +234,28 @@ class IndexVamana {
...
+    read_types(
+        ctx,
+        group_uri,
+        &feature_datatype,
+        &id_datatype,
+        &adjacency_row_index_datatype);
+
+    auto type =
+        std::tuple{feature_datatype, id_datatype, adjacency_row_index_datatype};
+    if (uri_dispatch_table.find(type) == uri_dispatch_table.end()) {
```
jparismorgan, line 253: TODO(paris): This should be `clear_history_dispatch_table` - will fix it in the follow-up so I don't need to re-run CI for this PR.

NikolaosPapailiou (approved):

PR #375, opened 2024-05-14
Add cloud taskgraph support to Python Vamana's query()

jparismorgan:
### What
Here we add cloud taskgraph support to Python Vamana's query(). We also start testing for `BATCH` Vamana ingestion.

### Testing
* Existing tests pass.
* Vamana now runs the same cloud test as IVF Flat and passes.


apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -111,11 +148,135 @@ def query_internal(
...
+            The name of the resource class to use ("standard" or "large"). Resource classes define maximum
+            limits for cpu and memory usage. Can only be used in REALTIME or BATCH mode.
+            Cannot be used alongside resources.
+            In REALTIME or BATCH mode if neither resource_class nor resources are provided,
+            we default to the "large" resource class.
+        resources:
+            A specification for the amount of resources to use when executing using TileDB cloud
+            taskgraphs, of the form: {"cpu": "6", "memory": "12Gi", "gpu": 1}. Can only be used
+            in BATCH mode. Cannot be used alongside resource_class.
+        num_workers: int
```
NikolaosPapailiou, line 203: This is not performing distributed computation with many workers. It is only running the centralized query execution using cloud and not client resources.

Lets not have placeholder args and documentation when the implementation is not yet in place.

NikolaosPapailiou (commented):
Adding support for query execution to be performed in a cloud taskgraph instead of the client could also be implemented in a generic way for all algorithms in `index.py` by wrapping `query_internal` and reopening the index inside the UDF.

apis/python/src/tiledb/vector_search/vamana_index.py
jparismorgan: Ah, yes, accident to include this, will remove.

jparismorgan:
Closed in favor of the new approach in #398

PR #383, opened 2024-05-16
Improve pytest latency

NikolaosPapailiou:
Remove 2 redundant tests to speedup Python CI latency.

The logic in these tests is also exercised in local mode. This test was mostly for cloud execution validation, which is already validated by the initial ingestion and update test.


jparismorgan:
30m 😍😍😍! Lmk if you are ready for review, would love to get this in!

jparismorgan (approved):

PR #388, opened 2024-05-16
Expose `build_config()` in Python

jparismorgan:
### What
When investigating an issue with Vamana cloud I wanted to print out information about the tiledb library version used in an UDF. You can do this to print out the tiledb Python library version:
```
>>> def testme():
...     import tiledb.vector_search
...     return tiledb.vector_search.version.version
...
>>> tiledb.cloud.udf.exec(testme, image_name="3.9-vectorsearch")
'0.4.0'
```
But I also noticed this function which was hidden behind a conditional directive. Here we start building it again, add a C++ test for it, and expose it in Python. This seems nice for potential future issues like this, though do let me know if you have concerns about exposing this to users.

### Result
This is what it prints out:
```
{"Build":"Release","Build_date":"2024-05-16 16:39:04","Compiler":{"CMAKE_CXX_FLAGS":"-Wall -Wno-unused-variable ","CMAKE_CXX_FLAGS_DEBUG":"-g","CMAKE_CXX_FLAGS_RELEASE":"-O3 -DNDEBUG","CMAKE_CXX_FLAGS_RELWITHDEBINFO":"-O2 -g -DNDEBUG","CXX_COMPILER":"/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++","CXX_COMPILER_ID":"AppleClang","CXX_VERSION":"15.0.0.15000309"},"GIT_REPO":{"GIT_BRANCH":"main","GIT_COMMIT_DATE":"2024-05-16","GIT_COMMIT_HASH":"9effb8bc","GIT_COMMIT_TIME":"10:26:42","GIT_REPO_NAME":"TileDB-Vector-Search-4","GIT_REPO_URL":""},"Run_date":"Thu May 16 16:41:20 2024","cmake_source_dir":"/Users/parismorgan/repo/TileDB-Vector-Search-4/src","tiledb_version":{"major":2,"minor":23,"patch":0},"uuid":"fdntoharwtynnkithnxcrdcy"}
```

### DIFF
--- a/src/include/stats.h
+++ b/src/include/stats.h
@@ -223,37 +223,28 @@ static auto dump_logs = [](std::string filename,
 };
 
 static auto build_config() {
-  std::string uuid_;
-  std::string date_;
-  std::size_t uuid_size_ = 24;
-
-  auto seed = std::random_device();
-  auto gen = std::mt19937(seed());
-  auto dis = std::uniform_int_distribution<int8_t>(97, 122);
-  uuid_.resize(uuid_size_);
-  std::generate(uuid_.begin(), uuid_.end(), [&] { return dis(gen); });
-
   // This is failing today, but could perhaps be added back in the future.
   // char host_[16];
   // if (int e = gethostname(host_, sizeof(host_))) {
   //   std::cerr << "truncated host name\n";
   //   strncpy(host_, "ghost", 15);
   // }
+
+  std::string date;
   {
     std::stringstream ss;
     std::time_t currentTime = std::time(nullptr);
     std::string dateString = std::ctime(&currentTime);
     dateString.erase(dateString.find('\n'));
     ss << dateString;
-    date_ = ss.str();
+    date = ss.str();
   }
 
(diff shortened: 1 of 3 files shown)


src/include/stats.h
```diff
@@ -222,40 +222,29 @@ static auto dump_logs = [](std::string filename,
...
 
   auto&& [major, minor, patch] = tiledb::version();
 
   json config = {
-      {"uuid", uuid_},
-      {"host", host_},
-      {"Program", program_name},
       {"Build_date", CURRENT_DATETIME},
-      {"Run_date", date_},
+      {"Run_date", date},
```
NikolaosPapailiou, line 247: Why do we have 2 dates? Do we need both of them?

jparismorgan: Not sure why, but I agree having the run date doesn't seem very useful (and is potentially even bad because it will make diff-ing the results harder), so will remove.

PR #397, opened 2024-05-31
Update `index.py`'s `query()` to set up a taskgraph for querying the updates array, adding query taskgraph support to Flat and Vamana

jparismorgan:


jparismorgan:
Closed in favor of the new approach in #398

PR #398, opened 2024-06-03
Add query with driver implementation

NikolaosPapailiou:
### What
Add an option to use a driver UDF which will spin up a cloud taskgraph. This is useful for both Vamana and Flat because you can now run queries in a TileDB cloud taskgraph. It is also useful for them (and IVF Flat) because if you use this mode you can access resources which you may not have local permissions to access.

Addresses SC-48813.

### Testing
* Adds new Vamana cloud tests
* Adds tests for the new `driver_mode`'s

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -36,11 +36,15 @@ class Index:
     timestamp: int or tuple(int)
         If int, open the index at a given timestamp.
         If tuple, open at the given start and end timestamps.
+    load_index_data: bool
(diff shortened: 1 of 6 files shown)


apis/python/src/tiledb/vector_search/index.py
```diff
@@ -35,11 +36,15 @@ class Index:
     timestamp: int or tuple(int)
         If int, open the index at a given timestamp.
         If tuple, open at the given start and end timestamps.
+    load_index_data: bool
```
NikolaosPapailiou, line 39: The functionality of `load_index_data` is overlapping a bit with `memory_budget` and it is not clear which to use for each usecase. Could we rename this to something more indicative of the functionality of this PR i.e. `open_for_remote_query_execution`?

jparismorgan (approved):

PR #404, opened 2024-06-03
Add Python IVF PQ Index

jparismorgan:
### What
Still in-progress IVF PQ index.

### DIFF
--- a/apis/python/test/test_object_index.py
+++ b/apis/python/test/test_object_index.py
@@ -11,7 +11,6 @@ from tiledb.vector_search.object_readers import ObjectPartition
 from tiledb.vector_search.object_readers import ObjectReader
 
 EMBED_DIM = 4
-INDEXES = ["FLAT", "IVF_FLAT", "VAMANA"]
 
 
 # TestEmbedding with vectors of EMBED_DIM size with all values being the id of the object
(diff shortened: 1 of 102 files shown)


apis/python/test/test_object_index.py
```diff
@@ -251,10 +251,16 @@ def test_object_index(tmp_path):
             index_type=index_type,
             object_reader=reader,
             embedding=embedding,
+            num_subspaces=4,
         )
 
         # Check initial ingestion
         index.update_index(partitions=10)
+
+        # TODO(SC-48908): Fix IVF_PQ with object index queries and remove.
```
NikolaosPapailiou, line 260: Is this failing?

NikolaosPapailiou (approved):
Great work! LGTM

apis/python/test/test_object_index.py
jparismorgan: Yep, this fails - we do get back valid IDs in `objects["external_id"]`, but they are in a different order than expected. So was going to fix this in a follow-up if that's okay with you?
```
distances, objects, metadata = index.query(
        {"object": np.array([[dim_id, dim_id, dim_id, dim_id]])},
        metadata_array_cond=f"test_attr >= {dim_id}",
        k=5,
        **query_kwargs,
    )
    assert np.array_equiv(
        objects["external_id"], np.array([v_id, v_id + 1, v_id + 2, v_id + 3, v_id + 4])
    )
```

PR #407, opened 2024-06-06
Initial scaling strategy

cainamisir:
Scaling resources linearly (with heuristic on using less cpu resources for smaller datasets - introduces at most 10 seconds on < 20mil vectors from initial tests - but could be from resource allocation.

Also scales partition size based on the baseline (20 mil vectors of 128 dimensions (int8)) keeping bytesize the same. Idea here was that we don't use less than int8 so there wouldn't be overhead from larger data types (that reduces total number of vectors)

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -2113,7 +2113,6 @@ def ingest(
                 ),
                 namespace=namespace,
             )
-            cpu_scale_factor = max(1, int(DEFAULT_PARTITION_BYTE_SIZE / size_in_bytes))
             threads = 16
 
             if acn:
@@ -2156,7 +2155,12 @@ def ingest(
             max_input_size,
             input_size,
         ):
-            cpu = str(min(max(min_resource_cpu, threads * input_size / max_input_size), max_resource_cpu))
+            cpu = str(
+                max(
+                    min(max_resource_cpu, threads * input_size / max_input_size),
+                    min_resource_cpu,
+                )
+            )
             ram = (
                 str(
                     max(
@@ -2174,10 +2178,14 @@ def ingest(
         # We can't set as default in the function due to the use of `str(threads)`
         # For consistency we then apply all defaults for resources here.
         if ingest_resources is None:
-            ingest_resources = scale_resources(2, 16, 2, 16, DEFAULT_PARTITION_BYTE_SIZE, size_in_bytes)
+            ingest_resources = scale_resources(
+                2, 16, 2, 16, DEFAULT_PARTITION_BYTE_SIZE, size_in_bytes
+            )
 
         if consolidate_partition_resources is None:
-            consolidate_partition_resources = scale_resources(2, 16, 2, 16, DEFAULT_PARTITION_BYTE_SIZE, size_in_bytes)
+            consolidate_partition_resources = scale_resources(
+                2, 16, 2, 16, DEFAULT_PARTITION_BYTE_SIZE, size_in_bytes
+            )
 
         if copy_centroids_resources is None:
             copy_centroids_resources = {"cpu": "1", "memory": "2Gi"}
@@ -2189,8 +2197,14 @@ def ingest(
             }
 
         if kmeans_resources is None:
-            kmeans_resources = scale_resources(4, 32, 2, 8, DEFAULT_KMEANS_BYTES_PER_SAMPLE, training_sample_size_in_bytes)
-
+            kmeans_resources = scale_resources(
+                4,
+                32,
+                2,
+                8,
+                DEFAULT_KMEANS_BYTES_PER_SAMPLE,
+                training_sample_size_in_bytes,
+            )
 
         if compute_new_centroids_resources is None:
             compute_new_centroids_resources = {
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -2105,6 +2105,9 @@ def ingest(
                 namespace=namespace,
             )
             threads = 16
+            if(size < input_vectors_per_work_item):
+                cpu_scale_factor = input_vectors_per_work_item/size
+                threads = max(4, int(threads / cpu_scale_factor))
             if acn:
                 kwargs["access_credentials_name"] = acn
         else:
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,31 +2140,60 @@ def create_ingestion_dag(
...
+                compute_new_centroids_resources = {"cpu": "1", "memory": "8Gi"}
+
+            if assign_points_and_partial_new_centroids_resources is None:
+                assign_points_and_partial_new_centroids_resources = {
+                    "cpu": str(threads),
+                    "memory": "12Gi",
+                }
+        else:
+            scale_factor = input_vectors_per_work_item/size
+            print("Scale factor: ", scale_factor)
```
NikolaosPapailiou, line 2172: Avoid printing to stdout, if you want this for debug purposes use `logger.debug` instead.

apis/python/src/tiledb/vector_search/ingestion.py
NikolaosPapailiou, line 1: You can format the file by running

```
pre-commit run --files apis/python/src/tiledb/vector_search/ingestion.py
```

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,31 +2140,60 @@ def create_ingestion_dag(
...
         )
 
-        # We can't set as default in the function due to the use of `str(threads)`
-        # For consistency we then apply all defaults for resources here.
-        if ingest_resources is None:
-            ingest_resources = {"cpu": str(threads), "memory": "16Gi"}
 
-        if consolidate_partition_resources is None:
-            consolidate_partition_resources = {"cpu": str(threads), "memory": "16Gi"}
+        if(size >= input_vectors_per_work_item):
```
NikolaosPapailiou, line 2144: Why do we need to have 2 cases here? I would suggest to merge the logic by using min, max computation

```
mem_based_on_size=...
ram = min(16, max(2, mem_based_on_size))
```

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,31 +2140,60 @@ def create_ingestion_dag(
...
+            if compute_new_centroids_resources is None:
+                compute_new_centroids_resources = {"cpu": "1", "memory": "8Gi"}
+
+            if assign_points_and_partial_new_centroids_resources is None:
+                assign_points_and_partial_new_centroids_resources = {
+                    "cpu": str(threads),
+                    "memory": "12Gi",
+                }
+        else:
+            scale_factor = input_vectors_per_work_item/size
```
NikolaosPapailiou, line 2171: This should be computed based on the vectors byte size. An alternative would be

```
16G_task_memory_overhead=6
32G_task_memory_overhead=12
ram = size*dimensions*type_byte_size*task_memory_overhead
````

File: apis/python/src/tiledb/vector_search/ingestion.py [hunk 4]
```diff
@@ -2661,8 +2693,9 @@ def consolidate_and_vacuum(
         logger.debug("Number of workers %d", workers)
 
         # Compute task parameters for main ingestion.
+        SCALED_VECTORS_PER_WORK_ITEM = int((VECTORS_PER_WORK_ITEM * 128) / dimensions / np.dtype(vector_type).itemsize)
```
NikolaosPapailiou, line 2696: This is a bit confusing, maybe we can set the default partition size in bytes instead.

apis/python/src/tiledb/vector_search/ingestion.py
jparismorgan: Instructions are also here: https://github.com/TileDB-Inc/TileDB-Vector-Search/blob/main/documentation/Building.md#formatting

cainamisir: Solved, thanks

apis/python/src/tiledb/vector_search/ingestion.py
cainamisir: Refactored, good call!

apis/python/src/tiledb/vector_search/ingestion.py
cainamisir: Created a map instead

apis/python/src/tiledb/vector_search/ingestion.py
cainamisir: Refactored

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,25 +2152,77 @@ def create_ingestion_dag(
...
         if random_sample_resources is None:
-            random_sample_resources = {"cpu": "2", "memory": "6Gi"}
+            random_sample_resources = {
+                "cpu": "2",
+                "memory": "6Gi",
+            }
 
         if kmeans_resources is None:
-            kmeans_resources = {"cpu": "8", "memory": "32Gi"}
+            kmeans_resources = {"cpu": "8", "memory": get_scaled_ram_gi_kmeans("32Gi")}
```
NikolaosPapailiou, line 2219: Why don't we scale down kmeans CPU for smaller inputs?

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,25 +2152,77 @@ def create_ingestion_dag(
...
+        def convert_base_to_int(base):
+            return int(base[:-2])
+
+        def ram_requirement_bytes(overhead):
+            return int(task_memory_overhead[overhead]) * this_batch_size_bytes
+
+        kmeans_requirement_bytes = kmeans_batch_size_bytes * 250
+
+        # Function to convert ram_requirement_bytes to ram string (approximates to nearest multiple of 2)
+        def get_scaled_ram_gi(base):
```
NikolaosPapailiou, line 2169: I think that the logic of `get_scaled_ram_gi, get_scaled_ram_gi_kmeans` is very complicated for what we want to achieve.

We could instead have a util function
```
def scale_resources(min_resource, max_resource, max_input_size, input_size):
    return max(
       min_resource, 
       min(
            max_resource, 
            int(max_resource * input_size / max_input_size)
       )
   )
``` 

And then use it to compute all resources:
```
ingest_cpu = scale_resources(2, 16, DEFAULT_PARTITION_BYTE_SIZE, size_in_bytes)
ingest_ram = scale_resources(2, 16, DEFAULT_PARTITION_BYTE_SIZE, size_in_bytes)
kmeans_cpu = scale_resources(2, 8, DEFAULT_KMEANS_BYTES_PER_SAMPLE, training_sample_size_in_bytes)
ingest_ram = scale_resources(2, 32, DEFAULT_KMEANS_BYTES_PER_SAMPLE, training_sample_size_in_bytes)
```

cainamisir: Could you take a look now?

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,25 +2147,70 @@ def create_ingestion_dag(
...
+            min_resource_ram,
+            max_resource_ram,
+            min_resource_cpu,
+            max_resource_cpu,
+            max_input_size,
+            input_size,
+        ):
+            cpu = str(
+                max(
+                    min(max_resource_cpu, threads * input_size / max_input_size),
```
NikolaosPapailiou, line 2160: This should be `max_resource_cpu* input_size / max_input_size`

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2137,25 +2147,70 @@ def create_ingestion_dag(
...
         if random_sample_resources is None:
-            random_sample_resources = {"cpu": "2", "memory": "6Gi"}
+            random_sample_resources = {
+                "cpu": "2",
+                "memory": "6Gi",
+            }
 
         if kmeans_resources is None:
-            kmeans_resources = {"cpu": "8", "memory": "32Gi"}
+            kmeans_resources = scale_resources(
```
NikolaosPapailiou, line 2200: I would prefer to have 2 separate `scale_resource` function calls for CPU and memory in order to be more clear which `min,max` values apply to which resource.

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2093,6 +2094,14 @@ def create_ingestion_dag(
         partial_index_resources: Optional[Mapping[str, Any]] = None,
     ) -> dag.DAG:
         kwargs = {}
+
+        # We compute the real size of the batch in bytes.
+        size_in_bytes = size * dimensions * np.dtype(vector_type).itemsize
+        logger.debug("Batch size in bytes: %d", size_in_bytes)
```
NikolaosPapailiou, line 2100: nit: 

```
logger.debug("Input size in bytes: %d", size_in_bytes)
```

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2093,6 +2094,14 @@ def create_ingestion_dag(
...
+        training_sample_size_in_bytes = (
+            training_sample_size * dimensions * np.dtype(vector_type).itemsize
+        )
+        logger.debug("Kmeans batch size in bytes: %d", training_sample_size_in_bytes)
```
NikolaosPapailiou, line 2104: nit:
```
logger.debug("Training sample size in bytes: %d", training_sample_size_in_bytes)
```

apis/python/src/tiledb/vector_search/ingestion.py
cainamisir: Changed it

NikolaosPapailiou (approved):
Thanks for the changes! LGTM

jparismorgan (approved):

PR #410, opened 2024-06-10
Replace `assert`s with exceptions in C++ code

jparismorgan:
### What
Today we use `assert`'s in many places, but when built for release we use `set(CMAKE_CXX_FLAGS_RELEASE "-Ofast -DNDEBUG " CACHE STRING "" FORCE)` and `-DNDEBUG` will compile out these checks. Thus, if there are cases where these `assert`'s can be hit (i.e. #389 is one such case), then users may get unexpected behavior which we don't see when testing things with debug builds.

As an immediate fix we replace `assert` with exceptions which should give better error logs when things go wrong. 

I did also consider adding a custom `should_not_happen_error` which we can use in cases like this, but for now this seems like a good first step.

### Testing
Existing tests pass.


NikolaosPapailiou (approved):
Thanks for doing this! I think this will remove some of the uncertainties that we currently have about problematic segmentation faults

PR #412, opened 2024-06-11
Add script to create an EC2 instance and run ann-benchmarks 

jparismorgan:
### What
Here we add a script to create an EC2 instance and run ann-benchmarks. 

### Testing
Here are the results of a run - we both stream logs to the console and also log to a file in case the console log gets cut off. We also download the HTML and image results.

[ann-benchmarks-logs.txt](https://github.com/user-attachments/files/15791825/ann-benchmarks-logs.txt)

sift-128-euclidean_10_euclidean-batch

### TODO
In follow-up's we will:
* Add Flat, Vamana, and IVF PQ.
* Add other vector search implementations so we can benchmark against them.
* Get this running in a CI job which we run on releases.


src/benchmarks/ann-benchmarks.py
```diff
@@ -0,0 +1,205 @@
...
+key_path = "/path/to/key.pem"
+
+if not os.path.exists(key_path):
+    raise FileNotFoundError(
+        f"Key file not found at {key_path}. Please set the correct path before running."
+    )
+
+# You do not need to change these.
+security_group_ids = ["sg-04258b401ce76d246"]
+instance_type = "r6i.16xlarge"
```
NikolaosPapailiou, line 31: Can you add a note on the resources of this `64CPU, 512GB`? 

This also seems quite large for 1M vectors. Are these resources suggested by `ann-benchmarks`? Should we also have a selection of instance types and be able to also run with less resources?

NikolaosPapailiou (approved):

src/benchmarks/ann-benchmarks.py
jparismorgan: Sure, will do. Yep, this is what they use, though they do so when running in single query mode with parallelism. So I guess we could use something smaller, I just thought matching would be simplest:

> These are all as of April 2023, running all benchmarks on a r6i.16xlarge machine on AWS with --parallelism 31 and hyperthreading disabled. All benchmarks are single-CPU.

PR #415, opened 2024-06-13
Improve benchmark script - add other vector search libraries and download full results

jparismorgan:
### What
Improve benchmark script:
- Add other vector search libraries.
- Download full results.
- Use env variable for your key name and path.
- Update to 30GB of storage b/c otherwise we would OOM when installing and running all the other vector search libraries.
- Update the results dir to store a new directory with the datetime in the path for each run.

### Testing
It was working, but something changed and now the tiledb docker install fails. I think it's still nice to move forward with getting this checked in, as once I fix that issue in `ann-benchmarks` it will work again.

### DIFF
--- a/src/benchmarks/ann-benchmarks.py
+++ b/src/benchmarks/ann-benchmarks.py
@@ -5,7 +5,7 @@
 # - Set up your AWS credentials locally. You can set them in `~/.aws/credentials` to be picked up automatically.
 # - Add environment variables which hold the values below. You can create these in the EC2 console.
 #   1. TILEDB_EC2_KEY_NAME: Your EC2 key pair name.
-#   2. TILEDB_EC2_KEY_PATH: The to your local private key file.
+#   2. TILEDB_EC2_KEY_PATH: The path to your local private key file.
 #     -  Make sure to `chmod 400 /path/to/key.pem` after download.
 # - caffeinate python src/benchmarks/ann-benchmarks.py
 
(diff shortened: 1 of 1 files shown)


src/benchmarks/ann-benchmarks.py
```diff
@@ -3,24 +3,63 @@
 # To run:
 # - pip install ".[benchmarks]"
 # - Set up your AWS credentials locally. You can set them in `~/.aws/credentials` to be picked up automatically.
-# - Fill in the following details. You can create these in the EC2 console.
-#   1. key_name: Your EC2 key pair name.
-#   2. key_path: The to your local private key file.
+# - Add environment variables which hold the values below. You can create these in the EC2 console.
+#   1. TILEDB_EC2_KEY_NAME: Your EC2 key pair name.
+#   2. TILEDB_EC2_KEY_PATH: The to your local private key file.
```
cainamisir, line 8: nit: probably missed "path" between 'the' and 'to'

cainamisir (approved):
Looks good! Just wondering why some of the benchmarks are failing?

jparismorgan:
> Looks good! Just wondering why some of the benchmarks are failing?

Some seem to be failing even in their CI ([sample run here](https://github.com/erikbern/ann-benchmarks/actions/runs/9543965090/job/26301636298)) and for others I believe we may need to pin specific Python packages, though not quite sure.

src/benchmarks/ann-benchmarks.py
jparismorgan: Thanks, fixed.

PR #419, opened 2024-06-18
Pin numpy to fix Python CI failures

NikolaosPapailiou:
This pins `numpy<2` as the new numpy release is causing Python CI failures


dudoslav (approved):
CI/CD seems to pass so looks good to me

PR #420, opened 2024-06-18
Fix IVF PQ time travel, use timestamp=1 as initial write timestamp from Python

jparismorgan:
### What
TODO

### Testing
TODO


jparismorgan:
Closing b/c instead of this we should fix IVF PQ to not write these arrays at timestamp=0.

PR #422, opened 2024-06-19
Distance metrics integration

cainamisir:
Added two new distance metrics to the FLAT index.

- Inner Product
- Cosine Distance

Usage: in create() or ingest(), distance_metric can be specified like this: 
`distance_metric=vspy.DistanceMetric.COSINE`
or
`distance_metric=vspy.DistanceMetric.INNER_PRODUCT`

Default distance metric is L2.

Note: the Inner Product distance metric returns -dot(a,b) as to provide the vectors in the order indicating highest similarity, without changing the paradigm that "distances" are returned in increasing order.


apis/python/src/tiledb/vector_search/flat_index.py
```diff
@@ -136,8 +136,20 @@ def query_internal(
                 (queries.shape[0], k), MAX_UINT64
             )
 
+        try:
+            distance_metric = vspy.DistanceMetric(
+                self.group.meta.get("distance_metric", vspy.DistanceMetric.L2)
+            )
+        except ValueError:
+            distance_metric = vspy.DistanceMetric.L2
+            raise ValueError(
```
NikolaosPapailiou, line 145: If you want to fallback to L2 as default you shouldn't use raise here.

Also `distance_metric` should only be configured during the index constructor.

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -250,6 +251,7 @@ def query(
         **kwargs
             Extra kwargs passed here are passed to the `query_internal` implementation of the concrete index class.
         """
+        # print all kwargs
```
NikolaosPapailiou, line 254: Is this a leftover comment? Please remove

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2755,6 +2758,11 @@ def consolidate_and_vacuum(
         )
         logger.debug("Number of workers %d", workers)
 
+        if distance_metric != vspy.DistanceMetric.L2 and index_type != "FLAT":
```
NikolaosPapailiou, line 2761: I would suggest to perform this check with the `create` function of each algorithm.

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2948,6 +2956,8 @@ def consolidate_and_vacuum(
             group.meta["partition_history"] = json.dumps(partition_history)
             group.meta["base_sizes"] = json.dumps(base_sizes)
             group.meta["ingestion_timestamps"] = json.dumps(ingestion_timestamps)
+            group.meta["distance_metric"] = int(distance_metric)
```
NikolaosPapailiou, line 2959: The code here is not called for all index algorithms. Move this inside the `create` function of each algorithm

NikolaosPapailiou:
Please add a bit more context for this change in the PR description and link to the respective shortcut issue.

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+
+    nn_cosine_sklearn = NearestNeighbors(n_neighbors=5, metric="cosine")
+    nn_cosine_sklearn.fit(dataset_vectors)
+    distances_sklearn, indices_sklearn = nn_cosine_sklearn.kneighbors(query_vectors)
+
+    indices_yourlib = index.query(query_vectors, k=5)
+
+    assert np.allclose(
+        distances_sklearn, indices_yourlib[0], 0.01
+    ), "Cosine similarity distances do not match"
```
jparismorgan, line 42: Can we lower `0.01`? Seems pretty big to me - if it's just about float precision, I've generally been able to use `1e-6` and have tests pass. But even a bit lower would be nice. Same below.

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,120 @@
...
+
+    dataset_vectors = load_fvecs(siftsmall_uri)
+    query_vectors = load_fvecs(queries_uri)
+
+    inner_products_sklearn = np.dot(query_vectors, dataset_vectors.T)
+
+    sorted_inner_products_sklearn = np.sort(inner_products_sklearn, axis=1)
+
+    results = index.query(query_vectors, k=5)
+    inner_products_yourlib = results[0]  # Assuming results are [distances, indices]
```
jparismorgan, line 66: (Minor) I don't think you need to say `Assuming` here, as this won't change, it's part of the API. But you could also do it directly with `distances, _ = index.query(query_vectors, k=5)`

apis/python/src/tiledb/vector_search/flat_index.py
cainamisir: Removed the fallback as it makes better sense to raise ther

apis/python/src/tiledb/vector_search/ivf_pq_index.py
```diff
@@ -177,6 +178,14 @@ def create(
...
             f"Number of dimensions ({dimensions}) must be divisible by num_subspaces ({num_subspaces})."
         )
+    if distance_metric != vspy.DistanceMetric.L2:
+        raise ValueError(
+            f"Distance metric {distance_metric} is not supported in IVF_PQ"
+        )
+
+    group = tiledb.Group(uri, "w")
+    group.meta["distance_metric"] = int(distance_metric)
+    group.close()
```
jparismorgan, line 188: We should do this in C++ instead of in Python for the type-erased indexes (IVF PQ and Vamana).

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    dataset_vectors = load_fvecs(siftsmall_uri)
+    query_vectors = load_fvecs(queries_uri)
+
+    nn_cosine_sklearn = NearestNeighbors(n_neighbors=5, metric="cosine")
+    nn_cosine_sklearn.fit(dataset_vectors)
+    distances_sklearn, indices_sklearn = nn_cosine_sklearn.kneighbors(query_vectors)
+
+    distances, indices = index.query(query_vectors, k=5)
```
jparismorgan, line 38: (Minor) Let's call these `ids`, not `indices` (you're right that by default they are indices, but b/c the user can set `external_ids`, we refer to them as `ids`)
```suggestion
    distances, ids = index.query(query_vectors, k=5)
```

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,118 @@
...
+
+    assert np.allclose(
+        distances_sklearn, distances, 1e-4
+    ), "Cosine similarity distances do not match"
+    assert np.array_equal(
+        indices_sklearn, indices
+    ), "Cosine similarity indices do not match"
+
+
+def test_inner_product(tmp_path):
```
jparismorgan, line 48: (Minor) We call these `test_cosine_similarity`, `test_inner_product`, and `test_l2_distance`. Shouldn't it be `test_cosine_distance` (b/c we're doing `1 - cosine similarity`) and `test_inner_product_distance`?

apis/python/src/tiledb/vector_search/ingestion.py
NikolaosPapailiou: Sure we could add it in `create_metadata`. However, we would also need to check wether each algorithm supports or not the specified distance

apis/python/src/tiledb/vector_search/flat_index.py
```diff
@@ -149,6 +151,7 @@ def create(
     group_exists: bool = False,
     config: Optional[Mapping[str, Any]] = None,
     storage_version: str = STORAGE_VERSION,
+    distance_metric: vspy.DistanceMetric = vspy.DistanceMetric.L2,
```
jparismorgan, line 154: (Minor) Let's add this to the docstring as well

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -59,6 +59,15 @@ def __init__(
...
         self.storage_version = self.group.meta.get("storage_version", "0.1")
+        try:
+            self.distance_metric = vspy.DistanceMetric(
+                self.group.meta.get("distance_metric", vspy.DistanceMetric.L2)
+            )
+        except ValueError:
+            raise ValueError(
+                f"Invalid distance metric in metadata: {self.group.meta.get('distance_metric')}."
+            )
+
```
jparismorgan, line 70: Just checking, is the `ValueError` that will be thrown here from an incorrect metadata value being set and the `vspy.DistanceMetric()` constructor throwing?

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -686,6 +696,7 @@ def _query_additions(
         nthreads=8,
         timestamp=None,
         config=None,
+        distance_metric=vspy.DistanceMetric.L2,
```
jparismorgan, line 699: Do we need a default here? Can we remove it and have this be a required argument? Seems nice to avoid defaults for private methods to prevent errors where someone forgets to pass `distance_metric`.

Same with `create_metadata()`, `declare_vq_query_heap()`, etc - basically if we do pass an argument to the function, let's not have a default.

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -191,6 +193,8 @@ def ingest(
         Resources to request when performing the write of centroids, only applies to BATCH mode
     partial_index_resources: Optional[Mapping[str, Any]]
         Resources to request when performing the computation of partial indexing, only applies to BATCH mode
+    distance_metric: Optional[vspy.DistanceMetric]
```
jparismorgan, line 196: (Minor)
```suggestion
    distance_metric: vspy.DistanceMetric
```

apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -199,6 +198,19 @@ static void declare_qv_query_heap_finite_ram(
...
+        //     parts_uri,
+        //     centroids,
+        //     query_vectors,
+        //     indices,
+        //     ids_uri,
+        //     nprobe,
+        //     k_nn,
+        //     upper_bound,
+        //     nthreads,
+        //     timestamp);
```
jparismorgan, line 212: (Minor) Let's remove please

apis/python/src/tiledb/vector_search/vamana_index.py
```diff
@@ -184,7 +185,12 @@ def create(
         dimensions=dimensions,
         l_build=l_build if l_build > 0 else L_BUILD_DEFAULT,
         r_max_degree=r_max_degree if l_build > 0 else R_MAX_DEGREE_DEFAULT,
+        distance_metric=int(distance_metric),
     )
+    if distance_metric != vspy.DistanceMetric.L2:
+        raise ValueError(
+            f"Distance metric {distance_metric} is not supported in VAMANA"
+        )
```
jparismorgan, line 193: Let's move this above `vspy.IndexVamana()`?

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,163 @@
...
+import pytest
+from array_paths import *
+from sklearn.neighbors import NearestNeighbors
+
+from tiledb.vector_search import _tiledbvspy as vspy
+from tiledb.vector_search.ingestion import ingest
+from tiledb.vector_search.utils import load_fvecs
+
+siftsmall_uri = siftsmall_inputs_file
+queries_uri = siftsmall_query_file
```
jparismorgan, line 13: (Minor) Can we not do this and instead reference `siftsmall_inputs_file` directly? It's what other tests do, so nice to be consistent.

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+def normalize_vectors(vectors):
+    norms = np.linalg.norm(vectors, axis=1, keepdims=True)
+    return vectors / norms
+
+
+def test_cosine_DISTANCE(tmp_path):
```
jparismorgan, line 21:
```suggestion
def test_cosine_distance(tmp_path):
```

apis/python/test/test_distance_metrics.py
```diff
@@ -0,0 +1,163 @@
...
+
+# def test_ivfpq_create_cosine(tmp_path):
+#     index_uri = os.path.join(tmp_path, "sift10k_flat_COSINE")
+#     with pytest.raises(RuntimeError):
+#         ingest(
+#             index_type="IVFPQ",
+#             index_uri=index_uri,
+#             source_uri=siftsmall_uri,
+#             source_type="FVEC",
+#             distance_metric=vspy.DistanceMetric.COSINE,
```
jparismorgan, line 162: Should these be commented back in?

src/include/api/ivf_pq_index.h
```diff
@@ -112,6 +113,19 @@ class IndexIVFPQ {
...
+            if (metric_value < 0 ||
+                metric_value > static_cast<int>(DistanceMetric::COSINE)) {
+              throw std::runtime_error(
+                  "Invalid distance metric value: " + value);
+            }
+            distance_metric = static_cast<DistanceMetric>(metric_value);
+          } catch (const std::exception& e) {
+            throw std::runtime_error(
+                "Error setting distance metric: " + std::string(e.what()));
+          }
```
jparismorgan, line 128: Can we make this into a helper function so we don't need to write it twice?

src/include/api/ivf_pq_index.h
```diff
@@ -562,22 +588,23 @@ class IndexIVFPQ {
   tiledb_datatype_t id_datatype_{TILEDB_ANY};
   tiledb_datatype_t partitioning_index_datatype_{TILEDB_ANY};
   std::unique_ptr<index_base> index_;
+  DistanceMetric distance_metric;
```
jparismorgan, line 591:
```suggestion
  DistanceMetric distance_metric_;
```
and also I think we need to define a default here, currently we'll have undefined behavior if the user doesn't pass it in the config. Would be good to add tests checking that this is set correctly a) when it's not explicitly set b) when it's explicitly set.

src/include/api/vamana_index.h
```diff
@@ -69,7 +70,6 @@ class IndexVamana {
   IndexVamana(IndexVamana&&) = default;
   IndexVamana& operator=(const IndexVamana&) = delete;
   IndexVamana& operator=(IndexVamana&&) = default;
-
```
jparismorgan, line 72: (Minor) There are a good amount of newline deletions / additions in this PR - are they on purpose? If so, that's fine with me, but if they are accident's then let's revert please.

src/include/index/ivf_pq_index.h
```diff
@@ -291,15 +293,17 @@ class ivf_pq_index {
       size_t max_iter = 2,
       float tol = 0.000025,
       std::optional<TemporalPolicy> temporal_policy = std::nullopt,
-      uint64_t seed = std::random_device{}())
+      uint64_t seed = std::random_device{}(),
+      DistanceMetric distance_metric = DistanceMetric::L2)
```
jparismorgan, line 297: (Minor) Probably best to put `distance_metric` before `seed` because it's much more common to set `distance_metric` than it is to set `seed`.

src/include/index/ivf_pq_index.h
```diff
@@ -1537,6 +1546,13 @@ class ivf_pq_index {
     if (num_clusters_ != rhs.num_clusters_) {
       return false;
     }
+    if (distance_metric_ != rhs.distance_metric_) {
+      std::cout << "distance_metric_ != rhs.distance_metric_"
+                << static_cast<int>(distance_metric_)
+                << " != " << static_cast<int>(rhs.distance_metric_)
+                << std::endl;
```
jparismorgan, line 1553: (Minor) I'm happy with adding printing here, but if we do, let's do it for all the cases above as well. So could be a separate small PR to improve this if you wanted (or it can be in this one). Same with the Vamana code below.

src/include/index/ivf_pq_metadata.h
```diff
@@ -99,6 +100,7 @@ class ivf_pq_metadata : public base_index_metadata<ivf_pq_metadata> {
       {"sub_dimensions", &sub_dimensions_, TILEDB_UINT32, true},
       {"bits_per_subspace", &bits_per_subspace_, TILEDB_UINT32, true},
       {"num_clusters", &num_clusters_, TILEDB_UINT32, true},
+      {"distance_metric", &distance_metric_, TILEDB_UINT32, false},
```
jparismorgan, line 103: I think for IVF PQ we can make this required as we don't have any indexes released yet:
```suggestion
      {"distance_metric", &distance_metric_, TILEDB_UINT32, true},
```
Though Vamana we'll need to leave it optional.

src/include/scoring.h
```diff
@@ -286,6 +286,45 @@ using inner_product_distance = _inner_product_distance::inner_product_distance;
...
+      throw std::runtime_error("Unsupported distance metric");
+  }
+}
+
+// Helper function to wrap the call to get_distance_function
+template <typename Func, typename... Args>
+auto call_distance_function(DistanceMetric metric, Args&&... args) {
+  return get_distance_function<Func>(metric)(std::forward<Args>(args)...);
+}
+
```
jparismorgan, line 327: Are these used? If not, let's remove.

jparismorgan (commented):
Could we also add C++ unit tests for this? Specifically that we can set and get the metadata correctly, both during index creation and after writing to a URI. Note that you can just add checks to existing tests, you probably don't need totally new test cases here.

apis/python/src/tiledb/vector_search/index.py
cainamisir: I see this pattern in a lot of places with stuff that was there before - like group_exists and config in create_metadata - are we sure we want to change this only for distance_metric?

jparismorgan: Yep we should change those, but, either way, we don't have defaults for the other 5 arguments in `create_metadata()`, so we can just follow that pattern instead.

src/include/index/ivf_pq_index.h
cainamisir: That would make us fail the backwards compatibility test unfortunately - Unless we want to change the test it has to come after seed (I think)

apis/python/src/tiledb/vector_search/index.py
cainamisir: Yep, true

src/include/api/ivf_pq_index.h
cainamisir: If you can, please take a look over what I tried to do now - I think it's pretty smooth and can be reused for future stuff

src/include/api/vamana_index.h
cainamisir: I think the pre formatter handles them?

jparismorgan: No, these line additions and removals are not automatically done, else CI wouldn't be passing on any other PRs..

src/include/index/ivf_pq_index.h
jparismorgan: Mmm that doesn't really make sense to me, can you explain more?

apis/python/src/tiledb/vector_search/index.py
jparismorgan: Got it, and is there any way for that to happen today? And what error does it throw before you catch and re-throw? 

I ask because I'm not totally convinced we need this error handling as it seems like something that cannot happen, and if it does, will get automatically thrown... (and the other metadata reads don't do this)

src/include/api/ivf_pq_index.h
jparismorgan: Lgtm thanks!

src/include/api/ivf_pq_index.h
```diff
@@ -114,6 +115,12 @@ class IndexIVFPQ {
           id_datatype_ = string_to_datatype(value);
         } else if (key == "partitioning_index_type") {
           partitioning_index_datatype_ = string_to_datatype(value);
+        } else if (key == "distance_metric") {
+          try {
+            distance_metric_ = parseAndValidateDistanceMetric(value);
+          } catch (const std::runtime_error& e) {
+            std::cerr << e.what() << std::endl;
+          }
```
jparismorgan, line 123: Why do we need `std::cerr << e.what() << std::endl;`? Can't we just let `parseAndValidateDistanceMetric(value)` throw?

jparismorgan (approved):

PR #428, opened 2024-07-01
Remove `b_backtrack` from Vamana index

jparismorgan:
### What
This is unused and we don't have a plan for it, so remove.

### Testing
Existing tests pass.


cainamisir (approved):
Cool

PR #430, opened 2024-07-01
Cleanups to `ivf_index()` C++ code, and small cleanups in Python

jparismorgan:
### What
While reading this code I made renames, added comments, and removed unused code to help my understanding. There should be no logic or behavior change here, just a no-op refactor.

### Testing
Tests pass.

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -1742,37 +1742,25 @@ def ingest(
             )
             if source_type == "TILEDB_ARRAY":
                 logger.debug("Start indexing")
-                if index_timestamp is None:
-                    ivf_index_tdb(
-                        dtype=vector_type,
-                        db_uri=source_uri,
-                        external_ids_uri=external_ids_uri,
-                        deleted_ids=StdVector_u64(updated_ids),
-                        centroids_uri=centroids_uri,
-                        parts_uri=partial_write_array_parts_uri,
-                        index_array_uri=partial_write_array_index_uri,
-                        id_uri=partial_write_array_ids_uri,
-                        start=part,
-                        end=part_end,
-                        nthreads=threads,
-                        config=config,
-                    )
-                else:
-                    ivf_index_tdb(
-                        dtype=vector_type,
-                        db_uri=source_uri,
-                        external_ids_uri=external_ids_uri,
-                        deleted_ids=StdVector_u64(updated_ids),
-                        centroids_uri=centroids_uri,
-                        parts_uri=partial_write_array_parts_uri,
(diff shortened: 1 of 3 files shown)


src/include/detail/ivf/index.h
```diff
@@ -5,7 +5,7 @@
  *
  * The MIT License
  *
- * @copyright Copyright (c) 2023 TileDB, Inc.
+ * @copyright Copyright (c) 2023 Tileinput_vectors, Inc.
```
cainamisir, line 8: nit: I think you replaced this by mistake

cainamisir (approved):
looks good!

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -1795,41 +1783,25 @@ def ingest_vectors_udf(
...
+                    index_array_uri=partial_write_array_index_uri,
+                    id_uri=partial_write_array_ids_uri,
+                    start=part,
+                    end=part_end,
+                    nthreads=threads,
+                    **(
+                        {"timestamp": index_timestamp}
+                        if index_timestamp is not None
+                        else {}
+                    ),
```
jparismorgan, line 1802: Just to share, beyond normal testing, I also tested this with a small example:
```
def foo(cool, timestamp = 1):
    print('timestamp', timestamp)

def test_foo():
    print()

    foo('a', timestamp=0)
    
    foo('a', 0)

    index_timestamp = None
    foo(
        'a', 
        **({"timestamp": index_timestamp} if index_timestamp is not None else {}
    )
    )

    index_timestamp = 2
    foo(
        'a', 
        **({"timestamp": index_timestamp} if index_timestamp is not None else {}
    )
    )
```
which prints:
```
timestamp 0
timestamp 0
timestamp 1
timestamp 2
```

src/include/detail/ivf/index.h
```diff
@@ -95,51 +95,54 @@ int ivf_index(
...
     // debug_matrix(indices, "indices");
 
     // Some variables for debugging
     // @todo remove these once we are confident in the code
-    auto mis = std::max_element(begin(indices), end(indices));
-    auto a = std::distance(begin(indices), mis);
-    auto b = std::distance(mis, end(indices));
-    auto misx = *mis;
+    // auto mis = std::max_element(begin(indices), end(indices));
+    // auto a = std::distance(begin(indices), mis);
```
NikolaosPapailiou, line 135: Why don't we remove these entirely?

NikolaosPapailiou (approved):

src/include/detail/ivf/index.h
jparismorgan: Sure, done, just was leaving b/c the comments somewhat suggested there was future work we wanted to do here with it.

PR #433, opened 2024-07-03
Remove `apis/python/requirements-py.txt`

jparismorgan:
### What
It seems we do not use this, so removing. It was also not used in the PR that it was introduced in: https://github.com/TileDB-Inc/TileDB-Vector-Search/commit/bdf72cd54a273644841659ad1927b20fbd253794

### Testing
CI still passes.


dudoslav (approved):
I think this file is not needed so feel free to delete it

PR #439, opened 2024-07-08
Remove use of `set_coords_filter_list` from dense array creation

jparismorgan:
### What
Dense arrays don't store coordinates/dimensions, thus you can't filter/compression them. Remove use of `set_coords_filter_list` here.

### Testing
Tests pass.


Shelnutt2 (approved):
Thanks!

NikolaosPapailiou (approved):

PR #440, opened 2024-07-10
Tune default ingestion configuration to avoid OOM errors

NikolaosPapailiou:
This is changing the default ingestion configurations to:
- Make sure that ingestion using default config doesn't fail with OOM in BATCH mode for different input configs
- Make sure that ingestion using default config doesn't fail with OOM when executed in a small local server (<16GB) 
- Reduce the default vector partition size to avoid large writes/reads that can cause I/O errors when using the REST server.
- Scale `kmeans_resources`, `training_sample_size`and `partitions` according to the expected kmeans algorithm complexity and not only the byte size of the sample vectors.

Tested this setup with multiple configurations of dimensions(128-1024), datatypes(uint8-float32), input sizes(1k-1B) and training sizes(<1M) and didn't get OOM errors.

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -313,11 +313,30 @@ def ingest(
         + "".join(random.choices(string.ascii_letters, k=10))
     )
     DEFAULT_ATTR_FILTERS = storage_formats[storage_version]["DEFAULT_ATTR_FILTERS"]
-    DEFAULT_PARTITION_BYTE_SIZE = 2560000000  # 2.5GB
+
+    # This is used to auto-configure `input_vectors_per_work_item`
+    # We set it to a size that based on testing allowed ingestion to complete without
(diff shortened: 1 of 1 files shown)


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -1407,7 +1426,7 @@ def update_centroids():
                 config=config,
                 verbose=verbose,
                 trace_id=trace_id,
-            )
+            ).astype(np.float32)
```
jparismorgan, line 1429: Is this needed even if we don't use sklearn?

jparismorgan (approved):

apis/python/src/tiledb/vector_search/ingestion.py
NikolaosPapailiou: Not sure, I think that our vector to centroids assignments could handle vectors being different type than the centroids. However we need to test this.

PR #447, opened 2024-07-15
Support IVF PQ consolidation by storing raw feature vectors and external IDs

jparismorgan:
### What
Here we update IVF PQ to store the raw feature vectors and their IDs. Doing this, and using the existing `shuffled_vectors` and `external_ids` array's, lets us hook into the logic for consolidation. 

It also fixes a bug in re-ingestion where we would call this and get back `vector_type = int8` because we would store the PQ-encoded vectors in `source_uri`:
```
      if source_type is None:
                source_type = autodetect_source_type(source_uri=source_uri)
            in_size, dimensions, vector_type = read_source_metadata(
                source_uri=source_uri, source_type=source_type
            )
```

### Interesting issue: `clear_history()`
Note that I did run into one interesting issue, specifically that I would get `TileDB internal: [Array::set_array_closed] May not perform simultaneous open or close operations.` when calling:
```
  // Here we open an up an index.
  auto index = IndexIVFPQ(ctx, index_uri);
  auto queries = ...;
  index.query(queries);

  // Here we call a static function to delete fragments in the index.
  IndexIVFPQ::clear_history(ctx, index_uri, 99);
```
which specifically comes from the calls to `tiledb::Array::delete_fragments()`:
```
src/include/index/ivf_pq_group.h

void clear_history_impl(uint64_t timestamp) {
    ....
    tiledb::Array::delete_fragments(
        cached_ctx_, pq_ivf_indices_uri(), 0, timestamp);
    tiledb::Array::delete_fragments(
        cached_ctx_, pq_ivf_ids_uri(), 0, timestamp);
    tiledb::Array::delete_fragments(
        cached_ctx_, pq_ivf_vectors_uri(), 0, timestamp);
```
This is happening because in `query_infinite_ram()` we call:
```
src/include/index/ivf_pq_index.h

  template <feature_vector_array Q>
  auto query_infinite_ram(const Q& query_vectors, size_t k_nn, size_t nprobe) {
    ...
    if (!partitioned_pq_vectors_ || ::num_vectors(*partitioned_pq_vectors_) == 0) {
      read_index_infinite();
    }
```
which calls:
```
src/include/index/ivf_pq_index.h

auto read_index_infinite() {
   ...
   partitioned_pq_vectors_ = std::make_unique<tdb_pq_storage_type>(
        group_->cached_ctx(),
        group_->pq_ivf_vectors_uri(),
        group_->pq_ivf_indices_uri(),
        group_->get_num_partitions() + 1,
        group_->pq_ivf_ids_uri(),
        infinite_parts,
        0,
        temporal_policy_);
```
which calls `tdbPartitionedMatrix`, where we open up Arrays for these URIs:
```
src/include/detail/linalg/tdb_partitioned_matrix.h

tdbPartitionedMatrix(
   ...
   , partitioned_vectors_array_(tiledb_helpers::open_array(
            tdb_func__,
            ctx_,
            partitioned_vectors_uri_,
            TILEDB_READ,
            temporal_policy))
   ...
   , partitioned_ids_array_(tiledb_helpers::open_array(
            tdb_func__,
            ctx_,
            partitioned_ids_uri_,
            TILEDB_READ,
            temporal_policy))
```
We then do not close this array even after `query_infinite_ram()` is complete, so we have open TileDB Array's.

I read https://docs.tiledb.com/main/background/internal-mechanics/concurrency#vacuuming, and while it is about vacuuming and not `tiledb::Array::delete_fragments()`, it seems to explain why Core throws an exception here:

Screenshot 2024-07-15 at 5 51 21 PM

So based on this I believe throwing the error is correct here and we should make sure that app code does not have a IVF PQ index open after query.

### Testing
* Existing tests pass
* Several existing tests which were commented out for IVF PQ because they failed are commented back in.

### DIFF
--- a/src/include/index/index_group.h
+++ b/src/include/index/index_group.h
@@ -347,7 +347,10 @@ class base_index_group {
   }
 
   /**
-   * @brief Clears all history that is <= timestamp.
+   * @brief Clears all history that is <= timestamp. Note that if this is called
+   * while another index is open, it will throw an error. This is a TileDB Core
+   * feature to prevent read operations from accessing fragments that are being
+   * deleted.
    */
   void clear_history(uint64_t timestamp) {
     if (opened_for_ != TILEDB_WRITE) {
@@ -359,8 +362,19 @@ class base_index_group {
     }
 
     metadata_.clear_history(timestamp);
-    tiledb::Array::delete_fragments(cached_ctx_, ids_uri(), 0, timestamp);
-    static_cast<group_type*>(this)->clear_history_impl(timestamp);
+    try {
+      tiledb::Array::delete_fragments(cached_ctx_, ids_uri(), 0, timestamp);
+      static_cast<group_type*>(this)->clear_history_impl(timestamp);
+    } catch (const tiledb::TileDBError& e) {
+      if (std::string(e.what()).find("simultaneous open or close operations") !=
+          std::string::npos) {
+        throw std::runtime_error(
+            "[index_group@clear_history] Cannot clear history because the "
+            "index is open. Make sure to close the index before clearing "
+            "history.");
+      }
+      throw e;
+    }
   }
 
   /**
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -463,10 +463,6 @@ class Index:
         """
         from tiledb.vector_search.ingestion import ingest
 
-        if self.index_type == "IVF_PQ":
-            # TODO(SC-48888): Fix consolidation for IVF_PQ.
-            raise ValueError("IVF_PQ indexes do not support consolidation yet.")
-
         fragments_info = tiledb.array_fragments(
             self.updates_array_uri, ctx=tiledb.Ctx(self.config)
         )
@@ -584,6 +580,7 @@ class Index:
                     vspy.IndexIVFPQ.clear_history(ctx, uri, timestamp)
                 else:
                     raise ValueError(f"Unsupported index_type: {index_type}")
+                group.close()
                 return
 
             ingestion_timestamps = [
--- a/apis/python/test/test_ingestion.py
+++ b/apis/python/test/test_ingestion.py
@@ -617,9 +617,6 @@ def test_ingestion_timetravel(tmp_path):
             timestamp=20,
         )
 
-        if index_type == "IVF_PQ":
-            # TODO(SC-48888): Fix consolidation for IVF_PQ.
-            continue
         index = index.consolidate_updates()
 
         # We still have no results before timestamp 10.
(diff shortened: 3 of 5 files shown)


src/include/index/index_group.h
```diff
@@ -359,8 +362,19 @@ class base_index_group {
     }
 
     metadata_.clear_history(timestamp);
-    tiledb::Array::delete_fragments(cached_ctx_, ids_uri(), 0, timestamp);
-    static_cast<group_type*>(this)->clear_history_impl(timestamp);
+    try {
+      tiledb::Array::delete_fragments(cached_ctx_, ids_uri(), 0, timestamp);
+      static_cast<group_type*>(this)->clear_history_impl(timestamp);
+    } catch (const tiledb::TileDBError& e) {
+      if (std::string(e.what()).find("simultaneous open or close operations") !=
```
NikolaosPapailiou, line 369: Do we actually keep the arrays open to avoid reopening them across many queries? Can we make sure that we close all arrays consistently especially for library function calls?

NikolaosPapailiou (approved):
I think it is nice to maintain the convention that `shuffled_vectors, shuffled_ids` are always having the source vectors and we use different arrays for algorithms that have lossy compression. At some point we would also want a configuration that could disable this for cases that users don't do updates and need to save the extra storage.

src/include/index/index_group.h
jparismorgan: So the issue is with `partitioned_pq_vectors_`, which is a  `tdbColMajorPartitionedMatrix`:
1. `query_finite_ram()`: here it looks like we actually have a bug because I don't see us clear `partitioned_pq_vectors_` after the query, which we should do so that we can query a second time.
2. `query_infinite_ram()`: here I don't think we want to reset `partitioned_pq_vectors_` because we've just loaded it, and by keeping it in memory we can make subsequent queries faster. Though of course that is at the cost of increased memory usage before that happens, but it seems worth it to me. Instead we could either a. Update `tdbColMajorPartitionedMatrix` to automatically close Array's when it has read all the data or b. Add a `tdbColMajorPartitionedMatrix::markFinishedLoading()` method. I'll plan to merge this, implement one of these in a follow-up (will start with a.), and remove the checks I added here.

NikolaosPapailiou: `query_infinite_ram` should have loaded everything in memory so there is no need to keep the arrays open. Finite ram queries can benefit from not re-opening the array but we don't seem to have that optimization yet.

jparismorgan: > query_infinite_ram should have loaded everything in memory so there is no need to keep the arrays open.
Yep, exactly. a. vs b. above are just how to implement that in `tdbColMajorPartitionedMatrix` - I'm trying with a. which will have `tdbColMajorPartitionedMatrix` automatically manage this.

> Finite ram queries can benefit from not re-opening the array but we don't seem to have that optimization yet.
True, they'd just need to reset internal bookkeeping / data but could avoid the close -> re-open. But not focusing on that here, though could be nice for the future.

jparismorgan: Okay have a PR here which implements approach a. to close Array's automatically: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/448

PR #448, opened 2024-07-16
`tdbPartitionedMatrix` will automatically close Array's when done reading

jparismorgan:
### What
Today `tdbPartitionedMatrix` will leave `std::unique_ptr<tiledb::Array> partitioned_vectors_array_` and `std::unique_ptr<tiledb::Array> partitioned_ids_array_` open until the destructor. This was causing issues in #447 because we have a static function on our indexes which will try to call `tiledb::Array::delete_fragments()` on these `Array`'s, but that will throw if the `Array`'s are already open somewhere, which they are in `tdbPartitionedMatrix`.

### Testing
* Existing unit tests pass.
* Undos changes to Python which were added in #447 to work around this issue.
* Adds new `clear history with an open index` unit tests to Vamana and IVFPQ which make sure you can query and then call `clear_history()`. Note that Vamana doesn't need the changes in this PR to pass that tests, but IVFPQ fails without it.
* Adds checks for `tdb_partioned_matrix.load() = false` in several places that didn't have it.

### DIFF
--- a/apis/python/test/test_ingestion.py
+++ b/apis/python/test/test_ingestion.py
@@ -717,9 +717,6 @@ def test_ingestion_timetravel(tmp_path):
                 second_num_edges = num_edges_history[1]
 
         # Clear all history at timestamp 19.
-        # With type-erased indexes, we cannot call clear_history() while the index is open because they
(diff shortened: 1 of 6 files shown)


NikolaosPapailiou (approved):
Thanks for fixing this!

PR #449, opened 2024-07-17
Avoid creating one temp array for each ingestion work item

NikolaosPapailiou:
This avoids the creation of one temp array per work item to hold the partition indices. Instead we use one array to hold the indices of all work items.

This reduces the start up cost of ingestion especially for `tiledb://` uris as we don't have to create and register many arrays.

To avoid doing a rolling release of first releasing the lib change and then using it in ingestion, I applied a temporary solution of writing the indices in a temporary local array and then copying to the main array after the library call. This will be removed after the next cloud release.


jparismorgan (approved):
Fine with me to leave a temporary fix in until we have a release.

PR #451, opened 2024-07-17
IVF_FLAT Distance metric

cainamisir:
This PR introduces support for the Cosine distance metric in IVF_FLAT.

Added distance_metric parameter to relevant functions in ingestion.py and ivf_flat_index.py.

Implemented vector normalization:
Added a new function normalize_source_vectors to normalize vectors during ingestion.
Updated ingest_vectors_udf, ingest_additions_udf, and other relevant functions to handle vector normalization during updates.

Modified the IVFFlatIndex class to support Cosine distance in query, update, and update_batch operations.
Updated the create function in ivf_flat_index.py to allow Cosine distance metric.

Vector normalization is performed during ingestion (a normalization node is added which other nodes depend on) and querying when Cosine distance is used.

Accuracy tests (recall) and correctness tests have been added as well.

Inner product is not yet supported.


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -1215,6 +1222,7 @@ def centralised_kmeans(
                         config=config,
                         verbose=verbose,
                         trace_id=trace_id,
+                        # normalized = True
```
NikolaosPapailiou, line 1225: Remove commented out code.

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -1229,6 +1237,9 @@ def centralised_kmeans(
                         trace_id=trace_id,
                     ).astype(np.float32)
 
+                # if distance_metric == vspy.DistanceMetric.COSINE:
```
NikolaosPapailiou, line 1240: Remove commented out code.

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2063,6 +2084,56 @@ def consolidate_partition_udf(
             parts_array.close()
             ids_array.close()
 
+    def normalize_vectors_and_read_write(
```
NikolaosPapailiou, line 2087: nit: change name to `normalize_vectors`

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2304,6 +2375,11 @@ def scale_resources(min_resource, max_resource, max_input_size, input_size):
             )
             return d
         elif index_type == "IVF_FLAT":
+            # write normalized vectors to uri
```
NikolaosPapailiou, line 2378: Remove commented out code.

apis/python/src/tiledb/vector_search/ivf_flat_index.py
NikolaosPapailiou, line 1: Lets add some exceptions for the distances that are not supported yet i.e. inner product

cainamisir: we do in the create function for ivf_flat

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2394,6 +2499,13 @@ def scale_resources(min_resource, max_resource, max_input_size, input_size):
                         raise ValueError(
                             f"The random sampling ran into an issue: num_sampled ({num_sampled}) != training_sample_size ({training_sample_size})"
                         )
+                # Add dependencies for normalization
+                for node in normalization_nodes:
+                    if copy_centroids_uri is not None:
+                        centroids_node.depends_on(node)
```
jparismorgan, line 2505: Can this actually happen?  From my reading it cannot b/c we're in the `else:` case, rigth?

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2394,6 +2499,13 @@ def scale_resources(min_resource, max_resource, max_input_size, input_size):
                         raise ValueError(
                             f"The random sampling ran into an issue: num_sampled ({num_sampled}) != training_sample_size ({training_sample_size})"
                         )
+                # Add dependencies for normalization
```
jparismorgan, line 2502: Do we need to do something like we do for `random_sample_node` below? What will happen right now if we  have `copy_centroids_uri = None`  and `training_sampling_policy != TrainingSamplingPolicy.RANDOM`? It seems like we won't have anything depend on `normalization_nodes`?

apis/python/src/tiledb/vector_search/ivf_flat_index.py
```diff
@@ -295,8 +302,41 @@ def query_internal(
...
+        if self.distance_metric == vspy.DistanceMetric.COSINE:
+            queries = normalize_vectors(queries)
+        return super().query(
+            queries,
+            k,
+            driver_mode,
+            driver_resources,
+            driver_access_credentials_name,
+            **kwargs,
+        )
```
jparismorgan, line 338: (Minor) This seems okay, but we could also just update these methods in `index.py` to have:
```
if self.index_type == "IVF_FLAT" and self.distance_metric == vspy.DistanceMetric.COSINE:
    vector = normalize_vector(vector)
```
Will be less code and make things easier to reason about.

apis/python/src/tiledb/vector_search/ivf_flat_index.py
```diff
@@ -519,12 +560,15 @@ def create(
...
-    if distance_metric != vspy.DistanceMetric.L2:
+    if (
+        distance_metric != vspy.DistanceMetric.L2
+        and distance_metric != vspy.DistanceMetric.COSINE
+    ):
         raise ValueError(
             f"Distance metric {distance_metric} is not supported in IVF_FLAT"
         )
 
+
```
jparismorgan, line 571: (Minor) Lots of whitespace changes in this PR, could be good to remove them.

apis/python/test/test_distance_metrics.py
```diff
@@ -1,32 +1,274 @@
...
 from tiledb.vector_search.ingestion import ingest
+from tiledb.vector_search.utils import MAX_UINT64
+from tiledb.vector_search.utils import is_type_erased_index
 from tiledb.vector_search.utils import load_fvecs
 
+siftsmall_uri = siftsmall_inputs_file
+queries_uri = siftsmall_query_file
+
+
+def check_default_metadata(
```
jparismorgan, line 26: Have a few comments on #460 which would be great if we could apply here as well

jparismorgan (commented):
Looking nice so far!

apis/python/src/tiledb/vector_search/ingestion.py
cainamisir: Just added a new dependency from the centroid nodes

apis/python/src/tiledb/vector_search/module.cc
```diff
@@ -165,16 +165,44 @@ static void declare_qv_query_heap_infinite_ram(
...
+              top_centroids,
+              mat,
+              query_vectors,
+              nprobe,
+              k_nn,
+              nthreads,
+              cosine_distance_normalized{});
+          return make_python_pair(std::move(r));
+        }
       },
```
jparismorgan, line 206: Heads up I'm seeing lots of new warnings like:
```
  /Users/parismorgan/repo/TileDB-Vector-Search/apis/python/src/tiledb/vector_search/module.cc:820:3: note: in instantiation of function template specialization '(anonymous namespace)::declare_vq_query_heap_pyarray<signed char, unsigned long long>' requested here
    declare_vq_query_heap_pyarray<int8_t>(m, "i8");
    ^
  /Users/parismorgan/repo/TileDB-Vector-Search/apis/python/src/tiledb/vector_search/module.cc:626:7: warning: non-void lambda does not return a value in all control paths [-Wreturn-type]
        });
```
when I build - seems  nice to add an `else()` here and to the rest of the functions that throws.

apis/python/src/tiledb/vector_search/ivf_flat_index.py
cainamisir: I am down to do it either way, just seemed like a clean way to do it since we are already using polymorphism

apis/python/test/test_distance_metrics.py
```diff
@@ -138,6 +331,84 @@ def test_vamana_create_cosine(tmp_path):
...
+            np.testing.assert_allclose(
+                manual_distance,
+                distances_l2[i][j],
+                rtol=1e-5,
+                atol=1e-8,
+                err_msg=f"L2 distance mismatch for query {i}, neighbor {j}",
+            )
+
+    # Clean up
+    tiledb.VFS()
```
jparismorgan, line 407: (Minor) Is this required? Can we remove this?

apis/python/test/test_distance_metrics.py
```diff
@@ -15,6 +21,193 @@ def normalize_vectors(vectors):
...
+    )
+    expected_distances = np.append(expected_distances, 0.0)
+
+    # delete the vector with external id 3
+    index.delete(external_id=3)
+
+    # consolidate the updates
+    index = index.consolidate_updates()
+
+    # Query the index
```
jparismorgan, line 200: (Minor) Could we also add this to make sure the order is correct even after an `update()` + `delete()`?
```
    distances, ids = index.query(
        queries=np.array([[2, 2, 2, 2]], dtype=np.float32), k=5
    )
    assert np.array_equal(ids, np.array([[5, 4, 2, 1, 0]], dtype=np.uint64))
    sorted_distances = np.sort(distances)
    assert np.allclose(distances, sorted_distances, 1e-4)
```

apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -2333,6 +2390,54 @@ def scale_resources(min_resource, max_resource, max_input_size, input_size):
...
+                                vector_type=vector_type,
+                                dimensions=dimensions,
+                                source_start_pos=start,
+                                source_end_pos=end,
+                                batch=input_vectors_per_work_item_during_sampling,
+                                config=config,
+                                verbose=verbose,
+                                trace_id=trace_id,
+                                name=f"normalize-vectors-{start}-{end}",
+                                resources=random_sample_resources,  # We can use similar resources as random sampling
```
jparismorgan, line 2431: (Minor) I think the code is self-explanatory enough?
```suggestion
                                resources=random_sample_resources,
```

jparismorgan (approved):
Looks great, nice work 🥳

apis/python/src/tiledb/vector_search/ivf_flat_index.py
jparismorgan: Sure, sounds good, your choice, I'm fine with either approach

apis/python/src/tiledb/vector_search/ingestion.py
jparismorgan: Thanks, looks good to me

PR #454, opened 2024-07-17
Update dimensions to be uint64_t in C++

jparismorgan:
### What
Right now in C++ we have dimensions as `size_t`, `uint64_t`, and `uint32_t` in different places. In the metadata of created Python indexes I see `dimensions` is `<class 'numpy.uint32'>`, and that is also what the C++ metadata saves it as:
```
std::vector<metadata_arithmetic_check_type> metadata_arithmetic_checks{
      // name, member_variable, type, required
      {"dimensions", &dimensions_, TILEDB_UINT32, false},
      ...
```

 So here we change all references of `dimensions` to be `uint32_t` in C++.

Alongside this we also change `num_subspaces_` and `sub_dimensions_` to `uint32_t` so that we can compute them using `dimensions_` without casting. This is also a bug fix because previously we had them as:
```
src/include/index/ivf_pq_index.h
  uint64_t num_subspaces_{0};
  uint64_t sub_dimensions_{0};
```
But would store them as:
```
src/include/index/ivf_pq_group.h
  {"num_subspaces", &num_subspaces_, TILEDB_UINT32, true},
  {"sub_dimensions", &sub_dimensions_, TILEDB_UINT32, true},
```

In a follow-up we'll continue these updates with our variables that also have bugs, but we will split this up into separate PRs to be easier to review.

### Testing
Existing tests pass.


NikolaosPapailiou (commented):
Can we make this `unit64_t` consistently? The max uint32 value is ~4B which is close to our large scale datasets O(1B)

jparismorgan:
> Can we make this `unit64_t` consistently? The max uint32 value is ~4B which is close to our large scale datasets O(1B)

Done! One impact is that our old C++ metadata uses uint32, so we now fail to read it. So here we make this a breaking change (and delete the existing backwards compatibility indexes). This seems okay as I don't believe any customers use Vamana and we haven't made a public announcement about it. Another benefit is that it would let us delete the old Vamana backwards compatibility indexes which didn't have compression and so are quite big (~60MB instead of 100KB).

I also tested that we can still open up indexes generated in C++ in Python:
```
(TileDB-Vector-Search-2) ~/repo/TileDB-Vector-Search-2 python backwards-compatibility-data/generate_data.py 0.8.0
true
(TileDB-Vector-Search-2) ~/repo/TileDB-Vector-Search-2 pytest apis/python/test/test_backwards_compatibility.py -s

========================================== test session starts ==========================================
rootdir: /Users/parismorgan/repo/TileDB-Vector-Search-2
plugins: nbmake-1.5.3, anyio-4.4.0
collected 1 item                                                                                        

apis/python/test/test_backwards_compatibility.py .

=========================================== 1 passed in 1.67s ===========================================
```

NikolaosPapailiou (approved):

PR #455, opened 2024-07-17
Fix C++ incorrect or ambiguous primitive types

jparismorgan:
### What
In much of the C++ code we use `auto` for functions and types. This has resulted in many variables where there are mismatches between `size_t`, `uint32_t`, and `uint64_t`, and we do implicit casts between them at different places. Here we remove the use of `auto` for (many, but not all) primitive types and plumb the correct types (as defined by the saved metadata) throughout the code.

We also update the metadata types we store on the index a bit. For most things we use `TILEDB_UINT64`, but we use `TILEDB_UINT32` in places where the number won't get big. I did think about whether we should just use `TILEDB_UINT64` everywhere for simplicity. So let me know if you like that more than mixing and matching based on what we store.

### Testing
Tests pass.


src/include/api/vamana_index.h
```diff
@@ -340,8 +340,8 @@ class IndexVamana {
         const std::string& storage_version) = 0;
 
     [[nodiscard]] virtual uint64_t dimensions() const = 0;
-    [[nodiscard]] virtual size_t l_build() const = 0;
-    [[nodiscard]] virtual size_t r_max_degree() const = 0;
+    [[nodiscard]] virtual uint64_t l_build() const = 0;
```
NikolaosPapailiou, line 343: Should this be uint32_t? Same bellow

src/include/index/flat_pq_index.h
```diff
@@ -78,12 +78,12 @@ class flat_pq_index {
  public:
   // metadata
   uint64_t dimensions_{0};
-  size_t num_subspaces_{0};
-  size_t sub_dimensions_{0};
-  size_t bits_per_subspace_{8};
-  size_t num_clusters_{256};
-  float tol_ = 0.001;
-  size_t max_iter_ = 16;
+  uint64_t num_subspaces_{0};
```
NikolaosPapailiou, line 81: Should `num_subspaces_, sub_dimensions_` be `uint32_t`?

NikolaosPapailiou (approved):

src/include/index/flat_pq_index.h
jparismorgan: Thanks, done.

src/include/api/vamana_index.h
jparismorgan: Thanks, sure, done.

NikolaosPapailiou (approved):

PR #457, opened 2024-07-23
Fix markdown format for storage spec

NikolaosPapailiou:
Fix markdown format for storage spec


jparismorgan (approved):
What command did you run to do this? Could we add it to `pre-commit`?

PR #461, opened 2024-07-24
Fix C++ warnings across codebase

jparismorgan:
### What
We have many C++ warnings which show up on CI, which makes it very hard to read through them to check for real issues. I.e. [here](https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/9601365845/job/26479765880#step:5:81) you can see a sample log with 2924 lines of warnings. Here we fix many of them and now have 2275 lines of warnings (there are still a lot left, but at least we've made some progress).

Beyond warnings, I did make a few changes where we before copied things, but now we keep references to them, i.e. changing `auto indices = partitioned_vectors.indices();` to `const auto &indices = partitioned_vectors.indices();`

### Testing
Existing tests pass.


NikolaosPapailiou (approved):
Thanks for this cleanup!

PR #463, opened 2024-07-25
Parallelize Vamana query

jparismorgan:
### What
We had code to run the Vamana query in parallel, but we were not using it. Here we update to use it.

### TODO
In the future we may want to allow users to limit how many threads to use, but for now it seems fine to use `hardware_concurrency`.

### Testing
Tests pass.

### Benchmarks
Before (note the y-axis goes to .35):

query_time_vs_accuracy

After (note the y-axis goes to .12):

query_time_vs_accuracy

So it seems that we got significantly faster (i.e. the last dot on the blue line is 5x faster), which is great.


NikolaosPapailiou (approved):

jparismorgan:
Okay I fixed the segfault in #474, so going to merge this now!

PR #468, opened 2024-07-29
Optimize distance computations for IVFPQ

NikolaosPapailiou:
Implemented a new query to quantized vectors distance implementation that is:
- Computing distance tables between each query and each pq_centroid
- Uses the query distance tables during query-vector distance computations

Tested the performance of all methods using SIFT-1M:
- asymmetric(previous default)
query_time_vs_accuracy
- symmetric: 
  - ~1,3x speedup, accuracy reduction
query_time_vs_accuracy
- new distance based on query to pq_centroid distance tables: 
  - more than 2x speedup, no accuracy reduction 
query_time_vs_accuracy

When comparing latency for a larger dataset of 60M vectors of 200 dimensions, the original latency was ~100s and after this it is 9s.

### DIFF
--- a/src/include/index/ivf_pq_index.h
+++ b/src/include/index/ivf_pq_index.h
@@ -1246,16 +1246,17 @@ class ivf_pq_index {
     auto&& [active_partitions, active_queries] =
         detail::ivf::partition_ivf_flat_index<indices_type>(
             flat_ivf_centroids_, query_vectors, nprobe, num_threads_);
+
+    auto query_vectors_encoded =
+        std::move(*pq_encode<Q, pq_ivf_centroid_storage_type>(query_vectors));
     return detail::ivf::query_infinite_ram(
         *partitioned_pq_vectors_,
         active_partitions,
-        query_vectors,
+        query_vectors_encoded,
         active_queries,
         k_nn,
         num_threads_,
-        make_pq_distance_asymmetric<
-            std::span<typename Q::value_type>,
-            decltype(pq_storage_type{}[0])>());
+        make_pq_distance_symmetric());
   }
 
   /**
--- a/src/include/index/ivf_pq_index.h
+++ b/src/include/index/ivf_pq_index.h
@@ -574,28 +574,41 @@ class ivf_pq_index {
     };
     return pq_distance{this};
   }
-
+  /**
+   * @brief Computes the distance between a query and and a pq_encoded_vector
+   * given the distance table of the query to the pq_centroids
+   * query_to_pq_centroid_distance_table.
+   *
+   * @param query_to_pq_centroid_distance_table Distance table of the query
+   * vector to the pq_centroids.
+   * @param pq_encoded_vector PQ encoded database vector.
+   *
+   */
   template <feature_vector U, feature_vector V>
-  float sub_distance_query_centroid_distances(const U& a, const V& b) const {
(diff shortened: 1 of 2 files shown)


src/include/index/ivf_pq_index.h
```diff
@@ -1246,15 +1303,17 @@ class ivf_pq_index {
...
         active_partitions,
-        query_vectors,
+        query_centroid_distances,
         active_queries,
         k_nn,
         num_threads_,
-        make_pq_distance_asymmetric<
-            std::span<typename Q::value_type>,
+        make_pq_distance_query_centroid_distances<
+            std::span<float>,
```
jparismorgan, line 1316: Curious, does `typename Q::value_type` not work anymore?

src/include/index/ivf_pq_index.h
```diff
@@ -813,6 +839,37 @@ class ivf_pq_index {
     return pq_vectors;
   }
 
+  template <
+      feature_vector_array U,
+      class Matrix,
+      class Distance = uncached_sub_sum_of_squares_distance>
+  auto pq_query_centroid_distances(
+      const U& training_set, Distance distance = Distance{}) const {
```
jparismorgan, line 847: (Minor)(Optional) Might be a bit easier to read if we just refer to these as `vectors`:
```suggestion
      const U& vectors, Distance distance = Distance{}) const {
```

src/include/index/ivf_pq_index.h
```diff
@@ -813,6 +839,37 @@ class ivf_pq_index {
     return pq_vectors;
   }
 
+  template <
+      feature_vector_array U,
+      class Matrix,
+      class Distance = uncached_sub_sum_of_squares_distance>
+  auto pq_query_centroid_distances(
```
jparismorgan, line 846: Might be nice to add a comment like:

> For each vector, we will iterate through each of it's subspaces, and for each subspace we will compute the distance from the vector's subspace to each of the (256) clusters we've created for that subspace. We then store that distance, effectively caching the distance computation. Because we've computed this, later on, when we have a another vector which has been PQ-encoded and we want to compute the distance from our vector to this PQ-encoded vector, all we need to do is, for each subspace:
> 1. Get the cluster which the PQ-encoded vector's subspace is closest to.
> 2. Look in our pq_query_centroid_distances matrix for that cluster for this subspace. The value is a distance which we add to our total distance.
> 3. Continue this for all subspaces.

Though please feel free to edit / move somewhere else.

jparismorgan (approved):
Looks great, thanks for adding the charts as well! 

Could you also add a TODO to `query_finite_ram()` to update to this? Or you could update `query_finite_ram()` in this PR if you'd like, though it's fine with me to do in a follow-up once we support it via Python (I can do it then if you'd prefer, no worries).

src/include/index/ivf_pq_index.h
NikolaosPapailiou: This is always float as it is the distance between the query and the centroid. It is not related to the query vector type anymore.

src/include/index/ivf_pq_index.h
NikolaosPapailiou: Added more documentation

src/include/index/ivf_pq_index.h
NikolaosPapailiou: Renamed to `query_vectors`

PR #469, opened 2024-07-30
Use single shared seed for rng generation in C++

jparismorgan:
### What
Previously we had several different ways in which we generated random numbers spread throughout the codebase. Here we take code from TileDB Core to have a single seed for the whole C++ codebase:
- https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/common/random/seeder.h
- https://github.com/TileDB-Inc/TileDB/blob/dev/tiledb/common/random/prng.h
- https://github.com/TileDB-Inc/TileDB/blob/dev/test/support/tdb_catch_prng.h

We set the seed in two ways:
- During Catch2 unit tests, we use `src/include/test/utils/set_seed_from_catch.h` to set the seed to the random value assigned by Catch2. This is nice b/c it helps uncover flaky tests / bugs in the code that a fixed seed may not catch.
- During all other cases, we fix the seed (to `1234`). This is nice for customers because then they get reproducible usage of the library - i.e. if they run the same query twice they will get the same result both times.

### Testing
- Adds new tests for `Seeder` and `PRNG`.
- I printed out and validated that during unit tests we are hitting `SetSeedFromCatch` and setting in on the singleton `Seeder`.
- I built with `pip install .` and then validated that we use `seed = 1234` in this case.
- I can use `set_seed(715199)` in `apis/python/test/local-benchmarks.py` and reproduce the segfault from #463 consistently. 
- Existing tests pass.


src/include/test/unit_flat_pq_index.cc
```diff
@@ -441,7 +441,7 @@ TEST_CASE(
 
   SECTION("pq_encoding") {
     auto avg_error = pq_idx.verify_pq_encoding(training_set);
-    CHECK(avg_error < 0.08);
+    CHECK(avg_error < 0.081);
```
NikolaosPapailiou, line 444: I don't think that the test values should be fine-tuned for the specific seed. I think that we should set generic expectations in the tests.

NikolaosPapailiou (approved):

src/include/test/unit_flat_pq_index.cc
jparismorgan: Yep, makes sense. I changed here though because previously we had a fixed seed - i.e. we used `0xdeadbeef`:
```
sub_kmeans_random_init(training_set, centroids_, sub_begin, sub_end, 0xdeadbeef);
```
And now we have a random seed. So it's more like we previously had tuned it perfectly for a single seed, and now it's random again.

PR #475, opened 2024-08-01
Add new test and cleanup C++ code

jparismorgan:
### What
Here we:
* Add a new test for `stdx::range_for_each`
* Add `const` to several places where we can
* Fix warnings about casts in several places

### Testing
Tests pass.


NikolaosPapailiou (approved):
Thanks for the cleanup!

PR #481, opened 2024-08-06
Add more testing for Vamana helpers, add option to skip top_k to improve ingestion speed

jparismorgan:
### What
Here we do two main things:
1. Add more testing for Vamana helper functions
2. Add an option to `best_first_O4()` to skip computing top_k ids and distances. This is useful in Vamana ingestion where they are unused.

### Performance
The accuracy numbers are the same and ingestion is sometimes faster, but sometimes not, so likely just variability in my local machine is causing the change.

**Charts**
Before

ingestion_time_vs_accuracy

After

ingestion_time_vs_accuracy

**Raw numbers**
Before
```
VAMANA_l_build=40_r_max_degree=10
  Ingestion (count: 1):
    Average Time: 181.4188 seconds
  Query (count: 5):
    Average Time: 2.4649 seconds
    Average Accuracy: 0.7436

VAMANA_l_build=40_r_max_degree=15
  Ingestion (count: 1):
    Average Time: 260.3602 seconds
  Query (count: 5):
    Average Time: 1.9147 seconds
    Average Accuracy: 0.8676

VAMANA_l_build=40_r_max_degree=20
  Ingestion (count: 1):
    Average Time: 341.2821 seconds
  Query (count: 5):
    Average Time: 2.0450 seconds
    Average Accuracy: 0.9201

VAMANA_l_build=40_r_max_degree=25
  Ingestion (count: 1):
    Average Time: 419.1305 seconds
  Query (count: 5):
    Average Time: 1.9463 seconds
    Average Accuracy: 0.9439

VAMANA_l_build=40_r_max_degree=30
  Ingestion (count: 1):
    Average Time: 456.7194 seconds
  Query (count: 5):
    Average Time: 2.0422 seconds
    Average Accuracy: 0.9562

VAMANA_l_build=40_r_max_degree=35
  Ingestion (count: 1):
    Average Time: 480.5166 seconds
  Query (count: 5):
    Average Time: 2.0756 seconds
    Average Accuracy: 0.9624

VAMANA_l_build=40_r_max_degree=40
  Ingestion (count: 1):
    Average Time: 503.7234 seconds
  Query (count: 5):
    Average Time: 2.1702 seconds
    Average Accuracy: 0.9662
```

After
```
VAMANA_l_build=40_r_max_degree=10
  Ingestion (count: 1):
    Average Time: 157.6301 seconds
  Query (count: 5):
    Average Time: 1.7090 seconds
    Average Accuracy: 0.7436

VAMANA_l_build=40_r_max_degree=15
  Ingestion (count: 1):
    Average Time: 239.2754 seconds
  Query (count: 5):
    Average Time: 1.9881 seconds
    Average Accuracy: 0.8676

VAMANA_l_build=40_r_max_degree=20
  Ingestion (count: 1):
    Average Time: 330.6436 seconds
  Query (count: 5):
    Average Time: 2.3247 seconds
    Average Accuracy: 0.9201

VAMANA_l_build=40_r_max_degree=25
  Ingestion (count: 1):
    Average Time: 410.9352 seconds
  Query (count: 5):
    Average Time: 2.1819 seconds
    Average Accuracy: 0.9439

VAMANA_l_build=40_r_max_degree=30
  Ingestion (count: 1):
    Average Time: 463.5973 seconds
  Query (count: 5):
    Average Time: 2.4142 seconds
    Average Accuracy: 0.9562

VAMANA_l_build=40_r_max_degree=35
  Ingestion (count: 1):
    Average Time: 500.3645 seconds
  Query (count: 5):
    Average Time: 2.4888 seconds
    Average Accuracy: 0.9624

VAMANA_l_build=40_r_max_degree=40
  Ingestion (count: 1):
    Average Time: 521.9184 seconds
  Query (count: 5):
    Average Time: 2.6005 seconds
    Average Accuracy: 0.9662
```

### Testing
Tests pass.

### Benchmarking
I also tested `best_first_O4` vs `best_first_O5` (with `std::cout << _timing_data.dump();`), and `best_first_O4` (which we use) seems to be faster:
```
auto best_first_O4 count: 5, sum: 20.04 µs, avg: 4.01 µs
auto best_first_O5: count: 5, sum: 40.71 µs, avg: 8.14 µs
```

### DIFF
--- a/src/include/detail/graph/best_first.h
+++ b/src/include/detail/graph/best_first.h
@@ -413,8 +413,6 @@ auto best_first_O3(
     // Extract vertex with minimum score from frontier
     auto [_, p_star] = frontier.front();
 
-    auto debug_p_star = p_star;
-
     auto p_star_iter = vertex_state_map.find(p_star);
     if (p_star_iter == vertex_state_map.end()) {
       throw std::runtime_error(
@@ -470,7 +468,6 @@ auto best_first_O3(
           continue;
         }
       }
-      auto debug_neighbor_state = neighbor_state_iter->second;
 
       score_type heuristic = distance(db[neighbor_id], query);
 
@@ -541,9 +538,6 @@ auto best_first_O4(
     uint32_t Lmax,
     bool skip_top_k = false,
     Distance&& distance = Distance{}) {
-  scoped_timer __{"[best_first_O4][1] outer"};
-  debug_vector(query, "query");
-
   using id_type = typename std::decay_t<Graph>::id_type;
   using score_type = float;
 
@@ -565,7 +559,7 @@ auto best_first_O4(
   // Will be returned from this function
   std::unordered_set<id_type> visited;
 
-  //  scoped_timer __{tdb_func__};
+  scoped_timer __{tdb_func__};
 
   score_type heuristic = distance(db[source], query);
   pq.insert(heuristic, source);
@@ -573,17 +567,11 @@ auto best_first_O4(
 
   id_type p_star = source;
   set_enfrontiered(vertex_state_property_map[p_star]);
-  // debug_vector(vertex_state_property_map, "vertex_state_property_map A");
   do {
-    scoped_timer _2{"[best_first_O4][2] do loop"};
-    // std::cout << "[best_first_O4] p_star: " << p_star << std::endl;
     visited.insert(p_star);
-    // debug_unordered_set(visited, "[best_first_O4] visited");
     set_visited(vertex_state_property_map[p_star]);
 
-    scoped_timer _2a{"[best_first_O4][2a] do loop"};
     for (auto&& [_, neighbor_id] : graph[p_star]) {
-      scoped_timer _3{"[best_first_O4][3a] for graph loop"};
       auto neighbor_state = vertex_state_property_map[neighbor_id];
       if (!is_unvisited(neighbor_state)) {
         continue;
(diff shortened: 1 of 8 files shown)


src/include/detail/graph/best_first.h
```diff
@@ -80,17 +78,12 @@ auto best_first_O0(
...
-        return true;
-      }
       pq.push({heuristic, neighbor_id});
       visited.insert(neighbor_id);
     }
   }
 
-  std::cout << "Goal not found: visited " << size(visited) << std::endl;
-  return false;
+  return visited;
```
NikolaosPapailiou, line 86: Can we avoid using `auto` return types in these functions? It is really confusing to change the return type of this function from `bool` to `set` and not understand if this will affect other parts of the code.

NikolaosPapailiou (commented):
I think the names of the functions here are really not descriptive. Can we avoid the O1-4 notation and use better names or at least document the difference between the different best first implementations?

src/include/detail/graph/best_first.h
jparismorgan: Sure, done here. Prefer to leave the others b/c they turn out to be very, very long, and my original plan for all of this was to just modify how we set up these templates: instead of `template <class Graph,`, we could have `template <typename id_type` and then specify the `Graph` data structure explicitly. This would make the return type much shorter and more understandable.

jparismorgan:
> I think the names of the functions here are really not descriptive. Can we avoid the O1-4 notation and use better names or at least document the difference between the different best first implementations?

I do want to do this, but would prefer to do it in a future refactor, likely alongside fixing `best_first_O2` and `best_first_O3`, which don't compile. And we can also do it for the functions in `src/include/detail/graph/greedy_search.h`. Is that okay with you?

NikolaosPapailiou (approved):
LGTM

PR #485, opened 2024-08-09
Add out-of-core query() support to IVF PQ

jparismorgan:
### What
Add out-of-core `query()` support to IVF PQ. We achieve this by limiting the number of vectors we load in at a single time. In Python the number of vectors is called `memory_budge`, in C++ it is `upper_bound`.

### Testing
* Adds C++ and Python unit tests for new functionality, including a test to `tdbPartitionedMatrix` to make sure we load all vectors correctly by the time `load()` is finished.
* Existing tests pass.


NikolaosPapailiou (approved):
Great change! Thanks for adding this!

PR #489, opened 2024-08-15
IVF_PQ: Remove unused pq_ivf_centroids, remove extra call to train_ivf(), add comments, cleanup code

jparismorgan:
### What
* Remove unused `pq_ivf_centroids`. I believe this was from Algorithm #3 mentioned in the code, but we don't currently use it, and I don't see a world in which we move to needing it, so remove. Note that the big change is that we no longer store it in the index when we write to URI. To me this seems okay because if we really need it again we can compute it when we load the index, we'd just PQ-encode the `flat_ivf_centroids_`, which is a matrix of `original vector dimensions x IVF partitions`.
* Remove extra call to `train_ivf()` in `train()`: This is duplicate work because we do it again in `add()`. In fact, I'd like to move shortly after this PR to a C++ API which just has `add()`, `write_index()`, and `query()`, because we don't need both `train()` and `add()`, we only need one.
* Add comments about different parts of the algorithm.
* Delete some other unused code.

### Testing
Tests pass.

### DIFF
--- a/src/include/index/ivf_pq_group.h
+++ b/src/include/index/ivf_pq_group.h
@@ -40,7 +40,7 @@
 #include "index/ivf_pq_metadata.h"
 
 // flat_ivf_centroids, partitioned_pq_vectors, partitioned_pq_index,
-// partitioned_pq_ids, cluster_centroids, distance_tables.
+// partitioned_pq_ids, cluster_centroids.
 
 [[maybe_unused]] static StorageFormat ivf_pq_storage_formats = {
     {"0.3",
@@ -53,8 +53,6 @@
          {"pq_ivf_indices_array_name", "partitioned_pq_vector_indexes"},
          {"pq_ivf_ids_array_name", "partitioned_pq_vector_ids"},
          {"pq_ivf_vectors_array_name", "partitioned_pq_vectors"},
-
-         {"distance_tables_array_name", "pq_symmetric_distance_tables"},
      }}};
 
 template <class index_type>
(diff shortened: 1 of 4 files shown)


src/include/index/ivf_pq_group.h
```diff
@@ -39,17 +39,15 @@
...
+// flat_ivf_centroids, partitioned_pq_vectors, partitioned_pq_index,
+// partitioned_pq_ids, cluster_centroids, distance_tables.
 
 [[maybe_unused]] static StorageFormat ivf_pq_storage_formats = {
     {"0.3",
      {
          // @todo Should these be kept consistent with ivf_flat?
          {"cluster_centroids_array_name", "pq_cluster_centroids"},
          {"flat_ivf_centroids_array_name", "uncompressed_centroids"},
-         {"pq_ivf_centroids_array_name", "partition_centroids"},
```
NikolaosPapailiou, line 52: We could also remove the distance tables as they are not used in the new query algorithm.

src/include/index/ivf_pq_index.h
```diff
@@ -774,34 +763,46 @@ class ivf_pq_index {
         training_set_ids.end(),
         feature_vectors_.ids());
 
-    auto num_unique_labels = ::num_vectors(flat_ivf_centroids_);
```
NikolaosPapailiou, line 777: Conceptually, I would prefer to move `train_pq, train_ivf` in `train` and have add only perform the pq vector encoding and vector shuffling part. I think this will be a better approach to move to a potential OOC ingestion implementation

src/include/index/ivf_pq_group.h
jparismorgan: Thanks, removed. Nice to simplify even more.

src/include/index/ivf_pq_index.h
jparismorgan: Sure, seems reasonable, though I'd prefer to leave for now and improve along with out-of-core work if that's okay with you.

NikolaosPapailiou: sgtm

NikolaosPapailiou (approved):

PR #490, opened 2024-08-15
Remove train() API from type-erased indexes

jparismorgan:
### What
Here we remove the `train()` API from type-erased indexes for a few reasons:
1. It is unused.
2. I don't see a path anytime soon with Vamana or IVF PQ to need this.
3. Removing it will make our code somewhat faster (just not needing to call into C++, even though I believe everything should be by reference).
4. The code is simpler.
5. If we want this API in the future, we can just add it back, but actually have it do something.

Note that I was motivated to do this because when looking at out of core ingestion, it seems easier to reason about things without having two separate APIs which do the same thing.

### Testing
Tests pass.

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -1636,13 +1636,12 @@ def ingest(
         )
         if index_type == "VAMANA":
             index = vspy.IndexVamana(ctx, index_group_uri)
-            index.train(data)
+            index.add(data)
         elif index_type == "IVF_PQ":
             index = vspy.IndexIVFPQ(ctx, index_group_uri)
-            index.train(data, partitions)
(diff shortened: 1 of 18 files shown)


NikolaosPapailiou (commented):
I think `train` is conceptually needed especially for OOC/distributed ingestion because it is the only operation that can be performed on a sample of the input and is needed before any `add` operation.

jparismorgan:
> I think `train` is conceptually needed especially for OOC/distributed ingestion because it is the only operation that can be performed on a sample of the input and is needed before any `add` operation.

Mmm so my reasoning was that we would remove it now b/c it's unused and adds unneeded complexity, and when we need it for distributed ingestion then we add it back (and it likely will look different then as well). But I'm happy to remove if we see we don't need it after adding OOC / distributed ingestion.

PR #491, opened 2024-08-15
Fix infinite loop in kmeans

jparismorgan:
### What
Before this change, if you called `kmeans_random_init()` with less vectors in `training_set` than `num_partitions_`, we would get into an infinite loop. Now we fill centroids in with zeroes instead.

We also add unit tests for this code as it was implicitly tested through other classes, but not explicitly, and not for edge cases like this.

### Testing
* Existing tests pass.
* New tests pass.


cainamisir (approved):
Looks good!

NikolaosPapailiou (approved):

PR #492, opened 2024-08-16
Update matrix to take in a `std::vector` instead of `std::initializer_list`

jparismorgan:
### What
`std::initializer_list` is not as useful as `std::vector` because you can only create it in the constructor. But during testing of an IVF PQ I wanted to use `std::vector` to construct a matrix in a for-loop. So here we change over. Performance is also not an issue here because we only use this constructor during tests.

### Testing
Tests pass.


cainamisir (approved):
This looks good to me - initializer list is after all just a wrapper

PR #501, opened 2024-08-22
Add scGPT and scvi embeddings and improve SOMA reader

NikolaosPapailiou:
- Add [scGPT](https://github.com/bowang-lab/scGPT) embedding implementation
- Add [scvi](https://docs.scvi-tools.org/en/1.0.0/tutorials/notebooks/api_overview.html) embedding implementation 
- Change the SOMA reader to pass more query arguments and return an AnnData object.
- Update the soma object search example notebook.


apis/python/src/tiledb/vector_search/object_readers/soma_reader.py
```diff
@@ -44,44 +48,56 @@ def __init__(
...
+        self.max_size = max_size
         self.timestamp = timestamp
-        context = tiledbsoma.SOMATileDBContext(tiledb_ctx=tiledb.Ctx(self.config))
-        exp = tiledbsoma.Experiment.open(self.uri, "r", context=context)
-        self.num_obs = exp.obs.count
-        self.obs_array_uri = exp.obs.uri
+
+        self.context = tiledbsoma.SOMATileDBContext(
+            tiledb_config=tiledb.default_ctx().config().dict()
+        )
```
jparismorgan, line 62: For my understanding: I see we got rid of `config: Optional[Mapping[str, Any]] = None,` and now just grab the default config from here. How are users supposed to set that?

NikolaosPapailiou: Embedding and reader init args are serialised and stored in group metadata. Therefore storing config and other runtime/sensitive data should be avoided. 

The convention is that the config should only be passed to `ObjectIndex` methods `__init__, create, update` that run reader and embedding methods inside 
```
with tiledb.scope_ctx(ctx_or_config=config):
```

Here, tiledbsoma doesn't work correctly with `tiledb.scope_ctx` and instead we get the default config from the context.

apis/python/src/tiledb/vector_search/object_readers/soma_reader.py
```diff
@@ -29,13 +29,17 @@ def id(self) -> int:
...
 
 
-# class SomaRNAXRowReader(ObjectReader):
-class SomaRNAXRowReader:
+class SomaAnnDataReader(ObjectReader):
     def __init__(
         self,
         uri: str,
+        measurement_name: str = "RNA",
+        X_name: str = "raw",
```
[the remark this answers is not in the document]
NikolaosPapailiou: This matches the `query.to_anndata` argument so I would prefer to maintain the argument naming.

apis/python/src/tiledb/vector_search/object_readers/soma_reader.py
Shelnutt2: Since context are not safe for multi-processing or serialization, we generally offer both the option to pass configurations at the top and manage the context creations inside of the classes, along with passing a context for reuse where possible. The `default_context` in tiledb-py can only be initialized once per process. I believe forcing the use of it here will restrict the user in the case if they wanted to access more than one dataset in the same session that might require different configurations.

NikolaosPapailiou: For Vector Search Python APIs we don't currently have `context` arguments but only `config` arguments that should be passed explicitly. 

The current convention for `ObjectReaders, ObjectEmbeddings` is that they should expect config to be set by `default_context`. Therefore, tiledb operations work without any config arguments. The `ObjectIndex` achieves that by always running reader and embedding code within:
```
with tiledb.scope_ctx(ctx_or_config=config):
```
For distributed execution, the `config` dictionary is serialized and not the `Context`. Each UDF is also setting the `scope_ctx`

apis/python/src/tiledb/vector_search/object_readers/soma_reader.py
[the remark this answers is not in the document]
jparismorgan: Sure, sounds good to me.

jparismorgan (approved):
`scvi` looks good to me.

PR #502, opened 2024-08-22
IVF_PQ re-ranking

jparismorgan:
### What
Adds support for re-ranking to IVF_PQ. This is controlled by `k_factor`. We first query for `k_factor * k` vector IDs and indices, then fetch the original (not PQ-encoded) vectors and re-rank those.

### Testing
* Adds several new tests

### Benchmarks
Here are the results with `local-benchmarks` SIFT 1 million. Note that this is with local benchmarks so doesn't include network latency. But for the _everything is local_ case we get a nice accuracy improvement with a fairly small speed penality:

IVF_PQ_query_time_vs_accuracy


apis/python/test/conftest.py
```diff
@@ -20,10 +20,10 @@ def no_output(capfd):
 
     # Fail if there is any output.
     out, err = capfd.readouterr()
-    if out or err:
-        pytest.fail(
-            f"Test failed because output was captured. out:\n{out}\nerr:\n{err}"
-        )
+    # if out or err:
```
NikolaosPapailiou, line 23: Should we uncomment this?

apis/python/test/test_index.py
```diff
@@ -344,6 +347,14 @@ def test_ivf_pq_index(tmp_path):
     query_and_check_distances(
         index, np.array([[2, 2, 2]], dtype=np.float32), 2, [[0, 3]], [[2, 1]]
     )
+    query_and_check_distances(
+        index,
+        np.array([[2, 2, 2]], dtype=np.float32),
+        2,
+        [[0, 3]],
+        [[2, 1]],
+        k_factor=2.0,
```
NikolaosPapailiou, line 356: Let's also add a test without specifying `k_factor` argument.

apis/python/test/test_ingestion.py
```diff
@@ -245,9 +245,17 @@ def test_ivf_pq_ingestion_u8(tmp_path):
...
+    query_accuracy = accuracy(result, gt_i)
+    assert query_accuracy > MINIMUM_ACCURACY and query_accuracy < 0.9
+
+    _, result = index_ram.query(
+        queries,
+        k=k,
+        k_factor=2,
+        nprobe=nprobe,
+    )
+    assert accuracy(result, gt_i) > 0.98
```
NikolaosPapailiou, line 258: The check of `k_factor=1 -> query_accuracy < 0.9` and `k_factor=2 -> query_accuracy > 0.98` is a bit brittle. I would suggest to just compare the 2 accuracies with each other and assert an improvement threshold.

src/include/index/ivf_pq_index.h
```diff
@@ -1244,16 +1281,141 @@ class ivf_pq_index {
...
+    return re_rank(
+        std::move(initial_distances),
+        std::move(initial_ids),
+        std::move(initial_indices),
+        query_vectors,
+        k_initial,
+        k_nn);
+  }
+
+  auto re_rank(
```
NikolaosPapailiou, line 1305: nit: I would prefer `rerank` or `rerank_results`

src/include/index/ivf_pq_index.h
```diff
@@ -1244,16 +1281,141 @@ class ivf_pq_index {
...
+      ColMajorMatrix<size_t>&& initial_indices,
+      const auto& query_vectors,
+      size_t k_initial,
+      size_t k_nn) {
+    if (k_initial == k_nn) {
+      return std::make_tuple(
+          std::move(initial_distances), std::move(initial_ids));
+    }
+
+    if (::num_vectors(feature_vectors_) == 0 && !group_) {
```
NikolaosPapailiou, line 1317: Should we also load all `feature_vectors_` for the infinite memory case during the constructor?

NikolaosPapailiou (commented):
Great work on implementing this! 

Can we also get the benchmark results for SIFT-1M or SIFT-10M and S3 storage? This will give us a good idea of whether we need more code optimizations on the reranking part of the code.

apis/python/test/test_index.py
jparismorgan: Mmm sorry, don't follow, can you explain more?

apis/python/test/conftest.py
jparismorgan: Oops, yes, thanks, done.

apis/python/test/test_ingestion.py
jparismorgan: Sure, not totally convinced it's any less brittle 😅, but updated.

src/include/index/ivf_pq_index.h
jparismorgan: Sure, done. Note `rerank` is not actually a word in the dictionary, but it does make it easier to write, and people know what it means anyways.

src/include/index/ivf_pq_index.h
jparismorgan: Yep so that will happen (we hit that code in the second call to `rerank_query()` in `rerank()`. 

But this was just to cover the case where you create an empty index and then try to query with re-ranking. But I realized we actually don't need to throw here, so updated by moving the check for `group_` below and only reading from the URI if we have the group. Also added a test for this case.

apis/python/test/test_index.py
jparismorgan: Ah, we already have that right above this though, so I think we're okay?

src/include/index/ivf_pq_index.h
NikolaosPapailiou: Sorry but I still don't understand where we load the full `feature_vectors_` array. Shouldn't this be inside `read_index_infinite`?

NikolaosPapailiou:
> > Can we also get the benchmark results for SIFT-1M or SIFT-10M and S3 storage? This will give us a good idea of whether we need more code optimizations on the reranking part of the code.
> 
> Yep, will do, just opened up #514 which will make this comparison easy. Would prefer to not block this PR on that though, if it's okay with you? The `k_factor = 1.f` case is unaffected by this, so seems fine to get it merged even if it does turn out to be slow for remote URIs.

As this is new functionality mostly relevant for larger datasets stored in object stores, I would prefer to see the performance of it to understand the current state. And since #514 is ready I think we should try to produce it.

src/include/index/ivf_pq_index.h
jparismorgan: Oh, oops, I thought you meant loading `feature_vectors_` during `add()`. You are correct, we do not load the full `feature_vectors_` anywhere except during `add()`.

In the constructor we don't know if we're opening for infinite or finite queries, that is per-query. We could perhaps load the full `feature_vectors_` during the call to `query_infinite()` (which calls `read_index_infinite()`), but then we will perhaps be loading way more vectors than we actually need to, and than we would if we just loaded them in `rerank()`. The benefit though would be we don't need to do any load during the second call to `query_infinite()`, whereas today we do.

NikolaosPapailiou: There are multiple options here, so maybe add some query/constructor parameters to make sure that we can leverage all of them and understand what happens in each case.

jparismorgan: Talked offline - will do this along with moving `memory_budget` into the index constructor in a follow-up.

jparismorgan:
Here are the results of `local-benchmarks.py` on SIFT 1M using `k=100`. It is run from my local laptop with an `index_uri` that is on AWS. 

`query_infinite_ram()`
query_time_vs_accuracy

`query_finite_ram()` with `memory_budget=1000000` for both IVF_FLAT and IVF_PQ:
query_time_vs_accuracy

Observations:
* IVF_PQ re-ranking seems to be a good improvement versus no re-ranking. With it, IVF_FLAT is still better, but IVF_PQ is now reasonably close.
* IVF_PQ infinite queries seem to do better than finite queries.

NikolaosPapailiou (approved):
I think we need to perform more benchmarks/optimizations to make sure that `IVF_PQ` is materially better than `IVF_FLAT` for large objectstore-based datasets and OOC query execution. However, I am ok to merge this as a first version of reranking and follow up in a more specific PR.

src/include/index/ivf_pq_index.h
NikolaosPapailiou: Please add a TODO or shortcut for this

jparismorgan: Started on it here! #518

PR #503, opened 2024-08-23
Add open method for Index class

NikolaosPapailiou:
This adds `open` method for `Index` class which can be used to open an index without knowing its `index_type`

Tested: unit-tests

### DIFF
--- a/apis/python/src/tiledb/vector_search/__init__.py
+++ b/apis/python/src/tiledb/vector_search/__init__.py
@@ -4,6 +4,7 @@ from tiledb.cloud.dag.mode import Mode
 from . import utils
 from .flat_index import FlatIndex
 from .index import Index
+from .index import open
 from .ingestion import ingest
 from .ivf_flat_index import IVFFlatIndex
 from .ivf_pq_index import IVFPQIndex
@@ -34,6 +35,7 @@ __all__ = [
     "VamanaIndex",
     "IVFPQIndex",
     "Mode",
+    "open",
     "load_as_array",
     "load_as_matrix",
     "ingest",
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -46,6 +46,7 @@ class FlatIndex(index.Index):
         config: Optional[Mapping[str, Any]] = None,
         timestamp=None,
         open_for_remote_query_execution: bool = False,
+        group: tiledb.Group = None,
         **kwargs,
     ):
         self.index_open_kwargs = {
@@ -60,6 +61,7 @@ class FlatIndex(index.Index):
             config=config,
             timestamp=timestamp,
             open_for_remote_query_execution=open_for_remote_query_execution,
+            group=group,
         )
         self._index = None
         self.db_uri = self.group[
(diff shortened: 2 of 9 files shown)


Shelnutt2 (changes requested):
Let's put together a proposal for this. I'd like to see consideration of just using the class constructor vs having a non-class member function for open. Having open on Index itself doesn't follow any existing pattern from other APIs and projects in TileDB.

NikolaosPapailiou:
> Let's put together a proposal for this. I'd like to see consideration of just using the class constructor vs having a non-class member function for open. Having open on Index itself doesn't follow any existing pattern from other APIs and projects in TileDB.

Can you explain why this is not a pattern used in other TileDB projects? How is this different that `tiledb.open` for `Arrays` or `tiledbsoma.Experiment.open` for single cell `Experiment`?

Shelnutt2:
> > Let's put together a proposal for this. I'd like to see consideration of just using the class constructor vs having a non-class member function for open. Having open on Index itself doesn't follow any existing pattern from other APIs and projects in TileDB.
> 
> Can you explain why this is not a pattern used in other TileDB projects? How is this different than `tiledb.open` for `Arrays` or `tiledbsoma.Experiment.open` for single cell `Experiment`?

Certainly. `tiledb.open` is a static function at the top level of a project, its not a static class member function. In SOMA, `tiledbsoma.open` is the preferred way and documented examples on opening the dataset. In TileDB-VCF we use `tiledbvcf.Dataset()` and handle it in a single constructor. There isn't a main method we propose that is a static class member for opening and decting the underlying type. Its important we aim for consistency across APIs so there is a general push to common UX. Either we should prompt the use of `Index` or a single top level open function.

NikolaosPapailiou:
> > > Let's put together a proposal for this. I'd like to see consideration of just using the class constructor vs having a non-class member function for open. Having open on Index itself doesn't follow any existing pattern from other APIs and projects in TileDB.
> > 
> > 
> > Can you explain why this is not a pattern used in other TileDB projects? How is this different than `tiledb.open` for `Arrays` or `tiledbsoma.Experiment.open` for single cell `Experiment`?
> 
> Certainly. `tiledb.open` is a static function at the top level of a project, its not a static class member function. In SOMA, `tiledbsoma.open` is the preferred way and documented examples on opening the dataset. In TileDB-VCF we use `tiledbvcf.Dataset()` and handle it in a single constructor. There isn't a main method we propose that is a static class member for opening and decting the underlying type. Its important we aim for consistency across APIs so there is a general push to common UX. Either we should prompt the use of `Index` or a single top level open function.

Moved `open` to the top level of the project.

Concerning using the `Index` constructor, at the moment `Index` is an abstract class with concrete subclasses for each algorithm. In my understanding, this pattern can only use a factory method to instantiate concrete subclasses and can't use the abstract class constructor directly. To use the `Index` constructor we would need to reorganize our existing class and hierarchy structure. Let me know if you have a solution in mind for this to work.

NikolaosPapailiou:
@Shelnutt2 is the current solution of `open` at the top level of the project consistent with other TileDB packages?

jparismorgan (approved):
Looks good to me and seems to simplify the user experience, though may want to confirm with Seth.

PR #507, opened 2024-09-02
Add `fixed_min_triplet_heap`

jparismorgan:
### What
Adds `fixed_min_triplet_heap` which can be used to store triplets instead of pairs. We also delete `threshold_heap` because it is unused.

Note that I did consider removing the `fixed_min_pair_heap` and switching everywhere over to `fixed_min_triplet_heap`, but this is a ton of code changes, including for IVF_FLAT Python code, and so this seems the easier path.

### Testing
* Adds all the same tests that we have for `fixed_min_pair_heap` but for `fixed_min_triplet_heap`
* Also adds several new tests for `fixed_min_triplet_heap`


src/include/utils/fixed_min_heap.h
```diff
@@ -175,7 +196,12 @@ class fixed_min_pair_heap : public std::vector<std::tuple<T, U>> {
       }
 
       // Insert, since there is room
-      Base::emplace_back(x, y);
+      if constexpr (std::tuple_size_v<Tuple> == 2) {
+        Base::emplace_back(x, y);
+      } else {
+        Base::emplace_back(x, y, z);
```
NikolaosPapailiou, line 202: Do we need checks for the case that we insert triples in heaps initialised for pairs or other such corner cases. Also add some tests about this.

NikolaosPapailiou (approved):

PR #518, opened 2024-09-06
Update IVF_PQ to set memory_budget in constructor, support preload feature_vectors and metadata only modes

jparismorgan:
### What
Makes three changes to IVF_PQ:
1. Update IVF_PQ to set `memory_budget` in constructor instead of during query. This matches what Python does and improves things because now the first infinite RAM query is not slower the rest.
2. Support a new mode where we preload feature_vectors for use in re-ranking. This is useful so that for smaller indexes we can load all the data we need (like faiss does).
3. Add a metadata-only open mode so that if the user sets `open_for_remote_query_execution` in Python we won't load data in C++, just metadata.

Also updates `ivf_pq_index_test` because it is flaky (here is a PR without these changes failing: https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/10809242895/job/29983933038?pr=520).

### Testing
* Updates tests to create infinite RAM and finite RAM indexes and make sure the results are the same.
* Existing tests pass.

### Benchmarks
Before this, the first IVF_PQ query with `query_infinite_ram()` when the index URI is a remote AWS URI would be slower than all the rest (because we first had to load the data with `read_index_infinite()`):

query_time_vs_accuracy

After this change, we instead get this (notice that with k_factor=1 each query is the same speed, but for others it is not because we are fetching the feature_vectors for re-ranking on each query):

query_time_vs_accuracy

And if we use `index = IVFPQIndex(index_uri, config, preload_k_factor_vectors=True)`, then we this (notice that we have pre-fetched the feature_vectors so there is no query time difference):

query_time_vs_accuracy

The code to run this is below:
```
def benchmark_ivf_pq():
    index_type = "IVF_PQ"
    timer = timer_manager.new_timer(index_type)

    k = 100
    queries = load_fvecs(SIFT_QUERIES_PATH)
    dimensions = queries.shape[1]
    gt_i, gt_d = get_groundtruth_ivec(SIFT_GROUNDTRUTH_PATH, k=k, nqueries=len(queries))

    for partitions in [200]:
        for num_subspaces in [dimensions / 4]:
            for k_factor in [1, 1.5, 2, 4, 8, 16]:
                tag = f"{index_type}_partitions={partitions}_num_subspaces={num_subspaces}_k_factor={k_factor}"
                logger.info(f"Running {tag}")

                index_uri = get_uri(tag)

                timer.start(tag, TimerMode.INGESTION)
                ingest(
                    index_type=index_type,
                    index_uri=index_uri,
                    source_uri=SIFT_BASE_PATH,
                    config=config,
                    partitions=partitions,
                    training_sampling_policy=TrainingSamplingPolicy.RANDOM,
                    num_subspaces=num_subspaces,
                )
                ingest_time = timer.stop(tag, TimerMode.INGESTION)

                # The index returned by ingest() automatically has memory_budget=1000000 set. Open
                # a fresh index so it's clear what config is being used.
                index = IVFPQIndex(index_uri, config)

                for nprobe in [10, 10, 15, 20, 25, 30]:
                    timer.start(tag, TimerMode.QUERY)
                    _, result = index.query(
                        queries, k=k, nprobe=nprobe, k_factor=k_factor
                    )
                    query_time = timer.stop(tag, TimerMode.QUERY)
                    acc = timer.accuracy(tag, accuracy(result, gt_i))
                    logger.info(
                        f"Finished {tag} with nprobe={nprobe}. Ingestion: {ingest_time:.4f}s. Query: {query_time:.4f}s. Accuracy: {acc:.4f}."
                    )

                cleanup_uri(index_uri)

    timer.save_and_print_results()
```

### DIFF
--- a/src/include/index/index_defs.h
+++ b/src/include/index/index_defs.h
@@ -59,13 +59,14 @@ enum class QueryType { FiniteRAM, InfiniteRAM };
 
 enum class IndexLoadStrategy {
   // Used by Python to support `open_for_remote_query_execution`. In this case
-  // we don't want to load
-  // any data, we just want to load metadata so that we can call
+  // we don't want to load any data, we just want to load metadata so that we
+  // can call
   // index.dimensions() and other metadata getters.
   ONLY_METADATA,
-  // Load the index data as well as metadata. This is the default behavior.
+  // Load the index data as well as metadata.
   DEFAULT,
-  // Here we preload the vectors for reranking.
+  // Load everything DEFAULT does, but also preload feature vectors for
+  // reranking.
   PRELOAD_VECTORS_FOR_RERANKING,
 };
 
--- a/src/include/index/index_defs.h
+++ b/src/include/index/index_defs.h
@@ -62,12 +62,12 @@ enum class IndexLoadStrategy {
   // we don't want to load any data, we just want to load metadata so that we
   // can call
   // index.dimensions() and other metadata getters.
-  ONLY_METADATA,
-  // Load the index data as well as metadata.
-  DEFAULT,
+  METADATA_ONLY,
+  // Load the PQ index data as well as metadata.
+  PQ_INDEX,
   // Load everything DEFAULT does, but also preload feature vectors for
   // reranking.
-  PRELOAD_VECTORS_FOR_RERANKING,
+  PQ_INDEX_AND_RERANKING_VECTORS,
 };
 
 /******************************************************************************
--- a/src/include/index/index_defs.h
+++ b/src/include/index/index_defs.h
@@ -58,8 +58,10 @@ enum class IndexKind { FlatL2, IVFFlat, Vamana, IVFPQ };
 enum class QueryType { FiniteRAM, InfiniteRAM };
 
 enum class IndexLoadStrategy {
-  // Used by Python to support `open_for_remote_query_execution`. In this case we don't want to load 
-  // any data, we just want to load metadata so that we can call index.dimensions() and other metadata getters.
+  // Used by Python to support `open_for_remote_query_execution`. In this case
+  // we don't want to load
+  // any data, we just want to load metadata so that we can call
+  // index.dimensions() and other metadata getters.
   ONLY_METADATA,
   // Load the index data as well as metadata. This is the default behavior.
   DEFAULT,
--- a/src/include/index/index_defs.h
+++ b/src/include/index/index_defs.h
@@ -57,6 +57,16 @@ enum class IndexKind { FlatL2, IVFFlat, Vamana, IVFPQ };
 
 enum class QueryType { FiniteRAM, InfiniteRAM };
 
+enum class IndexLoadStrategy {
+  // Used by Python to support `open_for_remote_query_execution`. In this case we don't want to load 
+  // any data, we just want to load metadata so that we can call index.dimensions() and other metadata getters.
+  ONLY_METADATA,
+  // Load the index data as well as metadata. This is the default behavior.
+  DEFAULT,
+  // Here we preload the vectors for reranking.
+  PRELOAD_VECTORS_FOR_RERANKING,
+};
+
 /******************************************************************************
  * Static info for arrays associated with an index group
  ******************************************************************************/
--- a/src/include/index/index_defs.h
+++ b/src/include/index/index_defs.h
@@ -58,11 +58,8 @@ enum class IndexKind { FlatL2, IVFFlat, Vamana, IVFPQ };
 enum class QueryType { FiniteRAM, InfiniteRAM };
 
 enum class IndexLoadStrategy {
-  // Used by Python to support `open_for_remote_query_execution`. In this case
-  // we don't want to load any data, we just want to load metadata so that we
-  // can call
-  // index.dimensions() and other metadata getters.
-  METADATA_ONLY,
+  // Will load the index as an OOC index.
+  PQ_OOC,
   // Load the PQ index data as well as metadata.
   PQ_INDEX,
   // Load everything DEFAULT does, but also preload feature vectors for
--- a/src/include/index/ivf_pq_index.h
+++ b/src/include/index/ivf_pq_index.h
@@ -188,6 +188,8 @@ class ivf_pq_index {
   /****************************************************************************
    * Index group information
    ****************************************************************************/
+  IndexLoadStrategy index_load_strategy_;
+  size_t upper_bound_;
   TemporalPolicy temporal_policy_;
   std::unique_ptr<ivf_pq_group<ivf_pq_index>> group_;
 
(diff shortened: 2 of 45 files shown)


src/include/index/index_defs.h
```diff
@@ -57,6 +57,19 @@ enum class IndexKind { FlatL2, IVFFlat, Vamana, IVFPQ };
...
 enum class QueryType { FiniteRAM, InfiniteRAM };
 
+enum class IndexLoadStrategy {
+  // Used by Python to support `open_for_remote_query_execution`. In this case
+  // we don't want to load any data, we just want to load metadata so that we
+  // can call
+  // index.dimensions() and other metadata getters.
+  ONLY_METADATA,
+  // Load the index data as well as metadata.
+  DEFAULT,
```
NikolaosPapailiou, line 67: I would prefer to use more descriptive names i.e. `METADATA_ONLY, PQ_INDEX, PQ_INDEX_AND_RERANKING_VECTORS`

src/include/index/ivf_pq_index.h
```diff
@@ -385,6 +405,56 @@ class ivf_pq_index {
             std::nullopt,
             num_clusters_,
             temporal_policy_);
+
+    if (index_load_strategy_ ==
```
NikolaosPapailiou, line 409: This is a bit confusing. Shouldn't this also have `index_load_strategy_ ==IndexLoadStrategy::DEFAULT ||`

jparismorgan: Thanks, so there are four cases:
* IndexLoadStrategy::PQ_INDEX, upper_bound = 0
  * Load partitioned_pq_vectors_
* IndexLoadStrategy::PQ_INDEX, upper_bound > 0
  * Do not load partitioned_pq_vectors_
* IndexLoadStrategy::PQ_INDEX_AND_RERANKING_VECTORS, upper_bound = 0
  * Load partitioned_pq_vectors_
* IndexLoadStrategy::PQ_INDEX_AND_RERANKING_VECTORS, upper_bound > 0
  * Throw an error because this is invalid

Before I had `if (index_load_strategy_ == IndexLoadStrategy::PQ_INDEX_AND_RERANKING_VECTORS || upper_bound == 0) {` which still worked, but we can actually simplify to `if (upper_bound == 0) {` and it will do the same thing. So updated, thanks.

src/include/index/ivf_pq_index.h
```diff
@@ -385,6 +405,56 @@ class ivf_pq_index {
...
+          infinite_parts,
+          0,
+          temporal_policy_);
+
+      partitioned_pq_vectors_->load();
+
+      if (::num_vectors(*partitioned_pq_vectors_) !=
+          size(partitioned_pq_vectors_->ids())) {
+        throw std::runtime_error(
+            "[ivf_flat_index@read_index_infinite] "
```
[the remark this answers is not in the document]
jparismorgan: Thanks, fixed.

src/include/index/ivf_pq_index.h
NikolaosPapailiou: To make this code/enum logic clearer, I would suggest to load data only based on `IndexLoadStrategy`. `upper_bound>0` should only be relevant/allowed for `IndexLoadStrategy::METADATA_ONLY`.

jparismorgan: I can if you really want, but I think that leads to a quite verbose API that won't be easy to use (if people start using the C++ API)...

NikolaosPapailiou: I am not sure why it is more verbose for C++ users. We already added the `IndexLoadStrategy` argument. And what I am saying is to reduce complexity by only allowing `upper_bound` only for `IndexLoadStrategy::METADATA_ONLY`. 

Atm  the different enums in that strategy are not well defined. For example, as you have it now, if a user sets `IndexLoadStrategy::METADATA_ONLY` the index is not really usable (can't perform queries), which is not a useful mode.

jparismorgan: > if a user sets IndexLoadStrategy::METADATA_ONLY the index is not really usable (can't perform queries), which is not a useful mode.

But that's the whole point - it's so that you open an index that will help you query for metadata and nothing else.

jparismorgan: But sure, I'll update

jparismorgan: Oh wait, doing that would actually make things quite complicated, as then this check won't work, we'll return early instead of constructing `flat_ivf_centroids_` and `cluster_centroids_`:
```
if (index_load_strategy_ == IndexLoadStrategy::METADATA_ONLY) {
      return;
    }

    flat_ivf_centroids_ =
        tdbPreLoadMatrix<flat_vector_feature_type, stdx::layout_left>(
            group_->cached_ctx(),
            group_->flat_ivf_centroids_uri(),
            std::nullopt,
            num_partitions_,
            0,
            temporal_policy_);

    cluster_centroids_ =
        tdbPreLoadMatrix<flat_vector_feature_type, stdx::layout_left>(
            group_->cached_ctx(),
            group_->cluster_centroids_uri(),
            std::nullopt,
            std::nullopt,
            num_clusters_,
            temporal_policy_);
```
I guess maybe the real issue here is around the definition of `METADATA_ONLY`...

NikolaosPapailiou: Yes the definition of `METADATA_ONLY` is the main problem here, maybe we should call it `OOC`. Opening an index without query capabilities is not really valid IMO.

jparismorgan: Well the point was to support Python's `open_for_remote_query_execution`:
```
enum class IndexLoadStrategy {
  // Used by Python to support `open_for_remote_query_execution`. In this case
  // we don't want to load any data, we just want to load metadata so that we
  // can call
  // index.dimensions() and other metadata getters.
  METADATA_ONLY,
  // Load the PQ index data as well as metadata.
  PQ_INDEX,
  // Load everything DEFAULT does, but also preload feature vectors for
  // reranking.
  PQ_INDEX_AND_RERANKING_VECTORS,
};
```

jparismorgan: But we could change to this if you want:
```
enum class IndexLoadStrategy {
  // Will load the index as an OOC index.
  PQ_OOC,
  // Load the PQ index data as well as metadata.
  PQ_INDEX,
  // Load everything DEFAULT does, but also preload feature vectors for
  // reranking.
  PQ_INDEX_AND_RERANKING_VECTORS,
};
```

jparismorgan: Okay, updated PR.

> I am not sure why it is more verbose for C++ users.
Well, you can see in the commit that to do an OOC query now requires more code than it did before: https://github.com/TileDB-Inc/TileDB-Vector-Search/pull/518/commits/bf8d4c00b1385fceb5dd96b80695b95d99d00dc3#diff-15e9e750c136c14df58a0721c199d0596c1513e422cd3aa33c0f40a5aeec2223L235

NikolaosPapailiou: IMO the `index_load_strategy` argument should be placed before `upper_bound` and default to `PQ_OOC`

jparismorgan: Sure, updated to move `index_load_strategy` before `upper_bound`.

NikolaosPapailiou (approved):

jparismorgan:
I'm going to force a merge here because the failures are unrelated to this PR and I'd like this PR in the upcoming release:
```
(209 durations < 0.005s hidden.  Use -vv to show these durations.)
=========================== short test summary info ============================
FAILED test/test_cloud.py::CloudTests::test_cloud_flat - ModuleNotFoundError: No module named 'tiledb.array'
FAILED test/test_cloud.py::CloudTests::test_cloud_ivf_flat - ModuleNotFoundError: No module named 'tiledb.array'
FAILED test/test_cloud.py::CloudTests::test_cloud_ivf_flat_random_sampling - ModuleNotFoundError: No module named 'tiledb.array'
=========== 3 failed, 94 passed, 657 warnings in 1233.55s (0:20:33) ============
```

PR #522, opened 2024-09-11
Fix bug for Index clear_history

NikolaosPapailiou:
We always need to have `0` as the first ingestion timestamp, even if we don't have respective write fragments in the base index arrays. If not, `updates` that are before the `first_ingestion_timestamp` will not be visible. 

Testing: unit tests


jparismorgan (approved):
Lgtm if tests are passing. But what is the context of this? Was this a reported bug? Or a new Core change?

NikolaosPapailiou:
> Lgtm if tests are passing. But what is the context of this? Was this a reported bug? Or a new Core change?

Uncovered this during writing the tutorial for clearing the history of an index.

PR #532, opened 2024-09-24
Add explicit markdown parser for `DirectoryTextReader`

NikolaosPapailiou:
Add explicit markdown parser for `DirectoryTextReader`

This is causing [failures](https://github.com/TileDB-Inc/TileDB-Documentation/actions/runs/11015378317/job/30588303124?pr=341) for the Academy index building CI task

### DIFF
--- a/apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
+++ b/apis/python/src/tiledb/vector_search/object_readers/directory_reader.py
@@ -355,6 +355,7 @@ class TileDBLoader:
             handlers={
                 "application/pdf": PyMuPDFParser(),
                 "text/plain": TextParser(),
+                "text/markdown": TextParser(),
                 "text/html": BS4HTMLParser(),
                 "application/msword": MsWordParser(),
                 "application/vnd.openxmlformats-officedocument.wordprocessingml.document": (
@@ -385,7 +386,7 @@ class TileDBLoader:
             mime_type = mimetypes.guess_type(self.uri)[0]
             f = vfs.open(self.uri)
 
-        if mime_type is None:
+        if mime_type is None or mime_type.startswith("text"):
             mime_type = "text/plain"
 
         if mime_type.startswith("image/"):


jparismorgan (approved):
LGTM. Non-blocking, but maybe CSV / JSON / Rich Text Format handlers would be nice as well?

BBC-Esq:
How is this used in the program?  I couldn't find any places where it's actually called.

NikolaosPapailiou:
> How is this used in the program? I couldn't find any places where it's actually called.

Here is an example of how to use the `DirectoryTextReader` https://github.com/TileDB-Inc/TileDB-Vector-Search/blob/main/apis/python/examples/object_api/text_search_documents.ipynb

BBC-Esq:
I see...so it's basically only there to extract text from various file formats.  You're using Langchain...What's the difference from my perspective of simply using this repo in conjunction with Langchain directly (i.e. not via this repo) instead?  Personal preference?

See example

https://github.com/BBC-Esq/VectorDB-Plugin-for-LM-Studio/blob/28c940bf03e2be185443400b66f2229d7a902667/src/document_processor.py#L10

https://github.com/BBC-Esq/VectorDB-Plugin-for-LM-Studio/blob/28c940bf03e2be185443400b66f2229d7a902667/src/constants.py#L819

Also, I have a suggestion.  The PymuPDF loader from Langchian is great.  It's faster than the pypdf2/pypdf based loader, pdfminer or whatever other ones they have...I've tested them all.  It's the fastest and equally if not more accurate.

With that being said, it blows goats in terms of how it extracts pages.  What I mean is, it only extracts one page at a time.  The pymupdf library itself does not have a functionality to extract all pages at once into a single body of text.  Nor does the langchain wrapper around it provide a way to, for example, extract all the pages and concatenate them.

When you use the pymupdf-based loader from Langchain each page of the pdf becomes its own "document object" as Langchain defines this phrase.  Let's say you use the "RecursiveCharacterTextSplitter." which I'm sure you know how it works.  It will take each "document object" - i.e. each page of a pdf - and chunk it accordingly.  This leads to two problems:

1) The maximum length of a chunk will be limited by the number of characters of a given page.  Newer embedding models support contexts much longer - e.g. https://huggingface.co/Alibaba-NLP/gte-large-en-v1.5

2) It disrupts the RecursiveCharacterTextSplitter or any other text splitter (even the regular one).  What any text splitter ideally needs is a single document object for each file "loaded" with all the text concantenated for that specific file.  That way the page breaks don't artificially break the text in improper places.  The only pdf "loader" from langchain that did it correctly, if I remember, was the pdfminer one, but it's slow as hell.

Here's a workaround I use that modifies langchain's parser class. It concatenates all pages into a single body of text before returning a document object.  this creates problems if you want "page" as part of the metadata.  you have to essentially chunk the text then retroactively trace back to which page of the pdf a chunk originated from.  I've done this in my program, but that's all I'll write for now.

Here's the modified pymupdf parser:

https://github.com/BBC-Esq/VectorDB-Plugin-for-LM-Studio/blob/main/src/Assets/pdf.py

NikolaosPapailiou:
> What's the difference from my perspective of simply using this repo in conjunction with Langchain directly (i.e. not via this repo) instead?

Our Object API offers utilities to handle object reading, embedding, vector indexing and vector search in one consolidated API which we believe has more features than the Langchain VectorDB API. It also offers a lot of utilities for distributed embedding, indexing and querying execution using the TileDB platform.

Concerning the use of different PDF parsers, it is possible to create an extension of our default `DirectoryTextReader` implementation and pass it as a reader class to you `ObjectIndex`. This can be done entirely on your side without any changes in this repo as long as you implement the `ObjectReader` interface.

PR #536, opened 2024-10-01
Disable consolidate_update_fragments for TileDB Cloud URIs

NikolaosPapailiou:
Disable `_consolidate_update_fragments` for TileDB Cloud URIs.

Fragment consolidation is not supported in TileDB Cloud.

### DIFF
--- a/apis/python/src/tiledb/vector_search/index.py
+++ b/apis/python/src/tiledb/vector_search/index.py
@@ -2,6 +2,7 @@ import concurrent.futures as futures
 import json
 import os
 import time
+import warnings
 from abc import ABCMeta
 from abc import abstractmethod
 from typing import Any, Mapping, Optional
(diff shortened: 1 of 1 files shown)


ypatia:
[sc-56613]

ypatia (approved):
LGTM if you checked that that's the only place in vector search repo that `tiledb.consolidate` can be called for `tiledb://` arrays.

apis/python/src/tiledb/vector_search/index.py
```diff
@@ -800,6 +800,10 @@ def _set_has_updates(self, has_updates: bool = True):
             self.group = tiledb.Group(self.uri, "r", ctx=tiledb.Ctx(self.config))
 
     def _consolidate_update_fragments(self):
+        # Disable update fragment consolidation for TileDB Cloud URIs
+        # as this is not supported.
+        if self.uri.startswith("tiledb://"):
```
ihnorton, line 805: IMO we should log a message here (once only) at a low threshold (debug level).

ihnorton (approved):

apis/python/src/tiledb/vector_search/index.py
NikolaosPapailiou: Added a warning for this

PR #539, opened 2024-10-01
Decouple consolidate and vacuum operations

NikolaosPapailiou:
Vacuum operations are not process safe and therefore we should avoid coupling them with consolidation operations in order to maintain thread/process safety for vector search updates

### DIFF
--- a/apis/python/src/tiledb/vector_search/flat_index.py
+++ b/apis/python/src/tiledb/vector_search/flat_index.py
@@ -145,6 +145,30 @@ class FlatIndex(index.Index):
 
         return np.transpose(np.array(d)), np.transpose(np.array(i))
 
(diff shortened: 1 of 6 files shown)


jparismorgan (approved):
Looks good, though we'll also need to do this for VAMANA and IVF_PQ at some point. Also curious if we could check that it worked correctly in a unit test, though if that's hard to do I'm okay to skip it.

NikolaosPapailiou:
> Looks good, though we'll also need to do this for VAMANA and IVF_PQ at some point. Also curious if we could check that it worked correctly in a unit test, though if that's hard to do I'm okay to skip it.

Sure we need to apply the same logic also in C++ for the new algorithms.
I added code to exercise this in the unit tests.

PR #543, opened 2024-10-11
Add ColPali embedding and support for processing PDFs as images

NikolaosPapailiou:


apis/python/src/tiledb/vector_search/embeddings/colpali_embedding.py
```diff
@@ -0,0 +1,102 @@
...
+    def load(self) -> None:
+        import torch
+        from colpali_engine.models import ColPali
+        from colpali_engine.models import ColPaliProcessor
+
+        if self.device is None:
+            if torch.cuda.is_available() and torch.cuda.device_count() > 0:
+                self.device = "cuda"
+            else:
+                self.device = "cpu"
```
jparismorgan, line 45: Maybe we could add `mps` as a third option? Likely not a super common use case, but I always appreciate if repos have this when I run pytorch things locally. And they do call it out here: https://github.com/illuin-tech/colpali/blob/7f53f53ed005cfbafd2640ca7705a6297a97971f/README.md?plain=1#L69

jparismorgan (approved):
Looks good, though also seems nice to add a unit test for the new code paths in `directory_reader.py`, if not too hard.

apis/python/src/tiledb/vector_search/embeddings/colpali_embedding.py
NikolaosPapailiou: Added

PR #548, opened 2024-10-14
Cleanup unit_ivf_pq_index.cc

jparismorgan:
### What
Cleanup unit_ivf_pq_index.cc.

### Testing
Passes.

### DIFF
--- a/src/include/api/ivf_pq_index.h
+++ b/src/include/api/ivf_pq_index.h
@@ -573,7 +573,7 @@ class IndexIVFPQ {
     }
 
     uint64_t nlist() const override {
-      return impl_index_.nlist();
+      return impl_index_.partitions();
     }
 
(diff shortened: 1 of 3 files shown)


src/include/api/ivf_pq_index.h
```diff
@@ -573,7 +573,7 @@ class IndexIVFPQ {
     }
 
     uint64_t nlist() const override {
```
NikolaosPapailiou, line 575: Should we also rename the API level function `nlist->partitions`?

NikolaosPapailiou (approved):

src/include/api/ivf_pq_index.h
jparismorgan: Sure, will do this in a follow-up PR!

PR #550, opened 2024-10-14
Update tdbMatrixMultiRange to support both slices and individual indices

jparismorgan:
### What
In the out-of-core IVF_PQ code we'll need to be able to read slices from a matrix. Here we update `tdbMatrixMultiRange` to support both reads with both slices and individual indices.

### Testing
Tests pass.

### DIFF
--- a/src/include/detail/linalg/tdb_matrix_multi_range.h
+++ b/src/include/detail/linalg/tdb_matrix_multi_range.h
@@ -104,7 +104,13 @@ class tdbBlockedMatrixMultiRange : public Matrix<T, LayoutPolicy, I> {
    * @param uri URI of the TileDB array to read.
    * @param indices The indices of the columns to read.
    * @param dimensions The number of dimensions in each vector.
-   * @param upper_bound The maximum number of vectors to read.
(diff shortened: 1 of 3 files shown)


src/include/detail/linalg/tdb_matrix_multi_range.h
```diff
@@ -109,23 +111,30 @@ class tdbBlockedMatrixMultiRange : public Matrix<T, LayoutPolicy, I> {
   tdbBlockedMatrixMultiRange(
       const tiledb::Context& ctx,
       const std::string& uri,
-      const std::vector<I>& column_indices,
       size_type dimensions,
+      QueryType query_type,
+      const std::vector<I>& column_indices,
```
NikolaosPapailiou, line 116: Please update the function doc to mention how the new args work.

NikolaosPapailiou (approved):

src/include/detail/linalg/tdb_matrix_multi_range.h
jparismorgan: Thanks, done.

PR #553, opened 2024-10-17
Add bioimage search utils and example notebook

NikolaosPapailiou:
Add bioimage search utils and example notebook

### DIFF
--- a/apis/python/src/tiledb/vector_search/embeddings/huggingface_auto_image_embedding.py
+++ b/apis/python/src/tiledb/vector_search/embeddings/huggingface_auto_image_embedding.py
@@ -1,0 +1,70 @@
+from typing import Dict, Optional, OrderedDict
+
+import numpy as np
+
+class HuggingfaceAutoImageEmbedding:
+    def __init__(
+        self,
+        model_name_or_path: str,
+        dimensions: int,
+        device: Optional[str] = None,
+        cache_folder: Optional[str] = None,
+        batch_size: int = 64,
+    ):
+        self.model_name_or_path = model_name_or_path
+        self.dim_num = dimensions
+        self.device = device
+        self.cache_folder = cache_folder
+        self.batch_size = batch_size
+        self.processor = None
+        self.model = None
+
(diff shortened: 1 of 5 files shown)


ktsitsi (commented):
Really great. One question how can we validate the similarity? In this type of images is not profound. In a test notebook I tried while ago the related_images included a top result the query image too. Validating that the `query image` is similar with the `top 1 related image`.

NikolaosPapailiou:
> Really great. One question how can we validate the similarity? In this type of images is not profound. In a test notebook I tried while ago the `related_images` included as a top result the `query` image too. Validating that the `query image` is similar with the `top 1 related image`.

In the example Notebook, I currently skip the top result as it should always be the query image selected. However, in actual use cases, the query image would be an image that is not in the dataset so this shouldn't be problem. About result validation, do you have anything else in mind? The result similarity is mostly based on the embedding selected. Maybe you can find a finetuned embedding model for biomedical images that could give better results.

apis/python/src/tiledb/vector_search/embeddings/huggingface_auto_image_embedding.py
```diff
@@ -0,0 +1,76 @@
...
+        embeddings = np.zeros((size, self.dim_num), dtype=np.float32)
+        for image_id in range(len(objects["image"])):
+            image_batch.append(
+                Image.fromarray(
+                    np.reshape(objects["image"][image_id], objects["shape"][image_id])
+                )
+            )
+            count += 1
+            if count >= self.batch_size:
+                print(image_id)
```
ktsitsi, line 61: Is this a leftover `print`?

ktsitsi (approved):
Misc

apis/python/src/tiledb/vector_search/embeddings/huggingface_auto_image_embedding.py
NikolaosPapailiou: Removed

jparismorgan (approved):

PR #565, opened 2025-02-11
[WIP] Support Current domain in ingestion code

kounelisagis:
[sc-62701]

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -412,12 +412,20 @@ def ingest(
         if source_type == "TILEDB_ARRAY":
             schema = tiledb.ArraySchema.load(source_uri)
             size = np.int64(schema.domain.dim(1).domain[1]) + 1
-            dimensions = np.int64(schema.domain.dim(0).domain[1]) + 1
+            # In the case that current domain is non-empty, we need to consider it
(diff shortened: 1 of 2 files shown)


apis/python/src/tiledb/vector_search/ingestion.py
```diff
@@ -412,12 +412,20 @@ def read_source_metadata(
         if source_type == "TILEDB_ARRAY":
             schema = tiledb.ArraySchema.load(source_uri)
             size = np.int64(schema.domain.dim(1).domain[1]) + 1
-            dimensions = np.int64(schema.domain.dim(0).domain[1]) + 1
+            # In the case that current domain is non-empty, we need to consider it
+            if hasattr(schema, "current_domain") and not schema.current_domain.is_empty:
+                dimensions = np.int64(schema.current_domain.dim(0).domain[1]) + 1
+            else:
+                dimensions = np.int64(schema.domain.dim(0).domain[1]) + 1
```
teo-tsirpanis, line 419: Because the number of dimensions is an immutable value that is selected at the time of creating the embeddings, it seems more appropriate to use the similarly immutable domain of TileDB arrays.

If we want to support ingesting a subset of dimensions from the input file, I think we should have the user explicitly specify it, rather than guessing it from sources like the current domain.

kounelisagis: I suppose this is outdated now.

Issue #567, opened 2025-04-02
Python 3.13 wheels


ivirshup:
Hey! 

We're trying to get python 3.13 support for cellxgene census and have [run into some trouble installing this package from source on CI](https://github.com/chanzuckerberg/cellxgene-census/actions/runs/14137623748/job/39868593004). Would it be possible to get wheels up on pypi for python 3.13?

Thanks!

ihnorton:
@ivirshup yes, modulo any build issues.

[sc-65269]

PR #569, opened 2025-04-03
Enable wheel production for python 3.13

dudoslav:
Enable Python 3.13 wheels for TileDB Vector Search.


ihnorton (approved):
Thanks! Could you please also add the cmake pin for the CI jobs?

ihnorton (approved):
Thanks!

kounelisagis:
Fixed #567

PR #571, opened 2025-04-04
Use `is_dir` instead of `dir_size` for non-existent folders after index deletion

kounelisagis:
Calling `dir_size` on non-existent paths causes `VFSException`s to be thrown according to: https://github.com/TileDB-Inc/TileDB/blob/c36bce6517a46ce75f9a7c9090bd833ad40521b8/tiledb/sm/filesystem/vfs.cc#L209-L212.
`is_dir` is more suitable to call after index deletions in tests.

This PR resolves the errors we're currently encountering in CI.

---

[sc-65377]


ihnorton (approved):
Thanks!

PR #574, opened 2025-04-08
Decrease wheel size

dudoslav:
Build wheels in Release mode with stripped binaries


kounelisagis:
Running this to check the wheel sizes after the change: https://github.com/TileDB-Inc/TileDB-Vector-Search/actions/runs/14469698569

kounelisagis (approved):
There is a significant size decrease. Thanks!

PR #575, opened 2025-04-08
Fix missing token CI test failures

kounelisagis:
This PR fixes the `tiledb.libtiledb.TileDBError: [TileDB::REST] Error: Cannot set curl auth; either token or username/password must be set.` error encountered in the cloud tests. The issue was that the rest token was exported after `pytest` was executed.

---

[sc-65473]


ypatia (approved):
Please add a description.

PR #580, opened 2025-08-06
fix(ingestion): Pad sparse matrices to correct dimensions

giacomo-ciro:
This commit fixes a bug in `ingestion.py` where sparse matrices with null entries at the end were being read with incorrect dimensions.

Previously, the `read_input_vectors` function would convert a sparse matrix to a dense matrix without accounting for trailing null columns. This could lead to a mismatch in dimensions later in the code.

For example, a sparse array whose schema specified a shape of`10x100` might be read as a `10x90` dense matrix if its last 10 columns were empty.

The fix adds a check to ensure the matrix is padded with zeros to the expected dimensions, preventing the dimension mismatch.

### DIFF
--- a/apis/python/src/tiledb/vector_search/ingestion.py
+++ b/apis/python/src/tiledb/vector_search/ingestion.py
@@ -941,7 +941,8 @@ def ingest(
             ) as src_array:
                 src_array_schema = src_array.schema
                 data = src_array[start_pos:end_pos, 0:dimensions]
-                return coo_matrix(
+                
+                matrix = coo_matrix(
                     (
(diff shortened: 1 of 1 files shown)


giacomo-ciro:
Amazing!

Thank you for taking the time to review and improve.

Sorry for the test, I didn't think about implementing it.

Congrats again for the amazing software you are developing, super happy to have helped a little :)

kounelisagis:
(There are some token-related error since this is your branch. Please ignore them.)

giacomo-ciro:
Perfect, let me know if there's something else you need from me

kounelisagis (approved):

PR #595, opened 2025-12-15
Fix backwards compatibility tests on Windows.

teo-tsirpanis:
On Windows, Git checks out text files with CRLF line endings. This can cause problems when reading the  backwards compatibility arrays, because TileDB supports only LF line endings at the moment in its line-delimited files.[^1]

This PR fixes the test failures, by instructing Git to treat all test data as `binary`, causing it to not modify line endings.

[^1]: In this case, the failures were caused by TileDB-Inc/TileDB#5662, but CRLF support has been broken in other places before that.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,3 @@
 .git_archival.txt  export-subst
 *.qmd linguist-language=RMarkdown
-backwards-compatibility-data/data/* binary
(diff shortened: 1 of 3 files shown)


kounelisagis (approved):
Thanks!

